PR #569, opened 2026-04-09
fix(api): distinguish decode-failed from not-cached in proposal diff

0xNeoDev:
## Problem

The `/versioned/proposals/:id/diff` endpoint returns the same 404 "Edit blob not cached" for two different cases:
1. The blob was never fetched from IPFS (row doesn't exist in `ipfs_cache`)
2. The blob was fetched but failed GRC-20 validation (row exists with `is_errored=true`, `data=NULL`)

This makes it impossible for frontend/callers to distinguish between a temporary caching gap (retry later) and a permanently broken edit (will never work).

## Fix

Query `is_errored` alongside `data` from `ipfs_cache` and return distinct responses:

| Case | Status | Response |
|---|---|---|
| No row in `ipfs_cache` | 404 | `{"error": "Not found", "message": "Edit blob not cached for this proposal"}` |
| Row exists, `is_errored=true` | **422** | `{"error": "Unprocessable", "message": "Edit blob failed GRC-20 validation and cannot be decoded", "uri": "ipfs://..."}` |
| Row exists, data present | 200 | Normal diff response |

## Context

This was discovered investigating proposal `7dd5982f...` at block 113387, which has a malformed GRC-20 edit (context index out of bounds). The blob exists on IPFS and was fetched by `hermes-ipfs-cache`, but the 404 response gave no indication that the issue was a decode failure rather than a missing cache entry.

## Test plan
- [x] Format, lint, type check pass
- [ ] Hit `/versioned/proposals/7dd5982f7d2e4320bdcd885f4bd78543/diff?spaceId=41e851610e13a19441c4d980f2f2ce6b` and verify 422 response with uri
- [ ] Hit a proposal with a valid cached blob and verify normal 200 diff response
- [ ] Hit a proposal with no cached blob and verify 404 response

### DIFF
--- a/api/src/versioned/__tests__/router.test.ts
+++ b/api/src/versioned/__tests__/router.test.ts
@@ -644,6 +644,69 @@ describe("GET /versioned/proposals/:id/diff", () => {
 			expect(body.message).not.toContain("QmTest123")
 		})
 
+		it("returns 422 when edit blob failed GRC-20 validation (is_errored=true)", async () => {
+			const proposalId = "00000000-0000-0000-0000-000000000001"
+			const spaceId = "00000000-0000-0000-0000-000000000002"
+			const now = Math.floor(Date.now() / 1000)
+
+			// Mock: getProposalWithPublishAction returns proposal with content_uri
+			db.execute.mockResolvedValueOnce({
+				rows: [
+					{
+						proposal_id: proposalId,
+						space_id: spaceId,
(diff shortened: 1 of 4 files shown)


0xyaco (approved):
@0xNeoDev should we add at least one unit test for this particular case?

0xNeoDev:
> @0xNeoDev should we add at least one unit test for this particular case?

Yes! ty!

PR #576, opened 2026-04-13
feat(monitoring): weekly image-pruner DaemonSet to reclaim node inodes

0xNeoDev:
## Summary
- Adds a DaemonSet (`monitoring/k8s/image-pruner.yaml`) that runs `crictl rmi --prune` on every node every Sunday night (Mon 06:00 UTC = Sun 22:00 PT) to reclaim inodes consumed by unreferenced container images.
- Logs structured BEFORE/AFTER `df -i /` and image counts (`event=before|after|summary|run_complete`) so each run is queryable in `kubectl logs` / Axiom.

## Why
DigitalOcean managed Kubernetes does not expose kubelet configuration, so we cannot tune the built-in image GC thresholds (which only watch disk **bytes**, not inodes). Images accumulated over 152 days of node lifetime pushed one node to 9.5% inodes free and triggered `NodeFilesystemFilesFillingUp`. A manual prune across all 5 nodes reclaimed ~20M inodes total — this DaemonSet automates that cleanup going forward.

## Design notes
- `DaemonSet` in `kube-system`. Internal clock-checking loop wakes every 10 minutes and runs once when `date -u +%u%H` matches `TARGET_DOW` + `TARGET_HOUR`. Pinned to Sunday night regardless of pod restart time.
- Privileged + `hostPID` so it can talk to containerd's socket (`/run/containerd/containerd.sock`) and inspect the host root fs via a read-only `hostPath` mount at `/host-root`.
- Tolerates all taints so it covers dedicated/tainted nodes too.
- `crictl` is pinned to `v1.33.0` via the `CRICTL_VERSION` env var — bump alongside the cluster's k8s minor version.
- Per-call timeout is 5m (`CRICTL_TIMEOUT`) because containerd rmi calls get slow under inode pressure; the default 2s caused `DeadlineExceeded` errors during the manual prune.
- Runs two prune passes — the first frees references, the second reaps orphaned layers.

## Test plan
- [x] `kubectl apply --dry-run=client -f monitoring/k8s/image-pruner.yaml` passes.
- [ ] Apply to the cluster (single shared cluster covers both staging and prod).
- [ ] Verify all 5 pods become Ready: `kubectl -n kube-system rollout status ds/image-pruner`.
- [ ] Force an immediate run on one pod to verify end-to-end: `kubectl -n kube-system exec <pod> -- crictl -t 5m rmi --prune` and inspect `kubectl logs`.
- [ ] (Optional) Temporarily override `TARGET_DOW`/`TARGET_HOUR` to the current UTC value to trigger a full scheduled run, then revert.
- [ ] Confirm logs flow into Axiom and `event=run_complete` lines are queryable.
- [ ] Monitor inode trend over the following week to confirm images aren't re-accumulating faster than pruned.

## Rollout
```bash
kubectl apply -f monitoring/k8s/image-pruner.yaml
kubectl -n kube-system rollout status ds/image-pruner
kubectl -n kube-system logs ds/image-pruner --tail=50
```

## Rollback
```bash
kubectl delete -f monitoring/k8s/image-pruner.yaml
```

### DIFF
--- a/.github/workflows/maintenance-deploy.yml
+++ b/.github/workflows/maintenance-deploy.yml
@@ -1,22 +1,22 @@
-name: Deploy Monitoring (Production)
+name: Deploy Maintenance (Production)
 
-# Deploys cluster-wide monitoring/maintenance manifests in monitoring/k8s/.
-# Uses `kubectl apply --prune` with the `managed-by=gha-monitoring-deploy`
+# Deploys cluster-wide maintenance manifests in maintenance/k8s/.
+# Uses `kubectl apply --prune` with the `managed-by=gha-maintenance-deploy`
 # selector so resources removed from the repo are also removed from the
 # cluster on the next deploy. Any new manifest must include the label
-# `managed-by: gha-monitoring-deploy` (on the resource AND on its pod
+# `managed-by: gha-maintenance-deploy` (on the resource AND on its pod
 # template if applicable) to be tracked.
 on:
   push:
     branches:
       - main
     paths:
-      - 'monitoring/k8s/**'
-      - '.github/workflows/monitoring-deploy.yml'
+      - 'maintenance/k8s/**'
+      - '.github/workflows/maintenance-deploy.yml'
   workflow_dispatch:
 
 concurrency:
-  group: monitoring-production
+  group: maintenance-production
   cancel-in-progress: false
 
 env:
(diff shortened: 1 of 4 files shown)


0xNeoDev:
Replaced by #577 (branch renamed from `feat/k8s-image-pruner-cronjob` → `feat/maintenance-image-pruner`, PR auto-closed when source branch was deleted).

PR #583, opened 2026-04-13
fix(api): include entities missing sorted property in ordered results

0xJagger:
Wait until Preston and Sophia answer slack thread


0xJagger:
Sophia confirmed the issue was already solved, so no need to implement this fix for the moment

PR #587, opened 2026-04-16
fix(api): cap pagination at every nesting level (root + nested)

0xNeoDev:
## Summary
- Extends Byron's pagination cap plugin (#412) to cover **nested connections** and the **omitted-`first` case** — both at every nesting level, not just the root.
- When a client sends a bare collection query like `{ entity(id: ...) { relationsList { ... } } }`, the plugin now **injects `first: 1000`** at every connection / simple-collection field so PostGraphile's SQL always has a `LIMIT` at every lateral.
- Previously, omitting `first` entirely bypassed the cap and returned unbounded results — a single `{ relations(filter: {toEntityId: ...}) { ... } }` call could return 75K+ rows (36.7MB). And even with the top-level cap, nested `entity(id:) { relationsList { ... } }` on a hub entity (e.g. Claim, 75K+ inbound relations) was uncapped because resolver-wrapping only reliably influences root-level SQL.

## What changed
- **Swapped `makeWrapResolversPlugin` for a `graphile-build` hook.** The old plugin used `makeWrapResolversPlugin`, which per PostGraphile docs only reliably influences SQL for root-level resolvers. Nested sub-collections (`Entity.relationsList`, `Entity.valuesList`, `Entity.backlinksList`) were uncapped — which is exactly the pattern geogenesis' `EntityPage` hits when a user opens a hub entity like Claim.
- **New implementation:** hook `GraphQLObjectType:fields:field:args` and register an `addArgDataGenerator` on every connection / simple-collection field. The generator's `pgQuery(queryBuilder)` runs inside SQL construction at the field's own nesting level, so a nested `relationsList` inside an `entity(id:)` selection is capped exactly like a top-level `relations` query.
- Two policies applied uniformly at every level:
  1. `assertPaginationWithinLimit(fieldArgs)` — throws `BAD_USER_INPUT` (400) if `first`/`last`/`offset` > `MAX_PAGINATION_LIMIT` (1000).
  2. If neither `first` nor `last` is supplied, `queryBuilder.first(MAX_PAGINATION_LIMIT)` — so bare sub-collections cannot resolve unbounded.
- Kept `applyDefaultFirstIfOmitted` and `assertPaginationWithinLimit` as pure exports so the existing unit tests still exercise the policy without standing up a full PostGraphile schema.

## Generated SQL (what this actually produces)

For a query like:
```graphql
{ entity(id: "claim") { relationsList { toEntity { valuesList { text } } } } }
```

Every lateral now has its own `LIMIT`:
```sql
SELECT ... FROM entity e WHERE e.id = 'claim'
LEFT JOIN LATERAL (
  SELECT ... FROM relations r WHERE r.to_entity_id = e.id
  LIMIT 1000                   -- ← from relationsList's pgQuery
) rel ON true
LEFT JOIN LATERAL (
  SELECT ... FROM entity te WHERE te.id = rel.from_entity_id
  LEFT JOIN LATERAL (
    SELECT ... FROM values v WHERE v.entity_id = te.id
    LIMIT 1000                 -- ← from valuesList's pgQuery
  ) val ON true
) toe ON true
```

## Behavior change
Clients that previously relied on omitting `first` to fetch every row — at root *or* nested — will now receive at most 1000 rows per level and must paginate.

This is intentional — aligns with the stated goal of #412: "reduce risk of expensive third-party queries overloading the API with unbounded pagination sizes." Oversized explicit values still get `BAD_USER_INPUT` rejects.

## One thing to keep in mind

The cap is **per-parent-row, not global.** A query that matches 500 parent entities with 1000 nested values each still pulls 500,000 rows total. If that ever becomes a problem, the next defense is query cost analysis (e.g. `@escape.tech/graphql-armor` cost-limit) — which multiplies `first * nested_first * nested_nested_first` and rejects over budget at validation time. But for the pattern we're fixing today (one hub entity with 75K relations), per-level `LIMIT` is exactly the right shape.

## Test plan
- [x] Unit tests for `applyDefaultFirstIfOmitted` (missing first, explicit first, last-only, null-input edge cases)
- [x] Unit tests for `assertPaginationWithinLimit` (first/last/offset at and above cap)
- [x] Seeded integration test: inserts 1005 entities, asserts bare `{ entities { id } }` returns exactly 1000
- [x] Same assertion via connection form (`entitiesConnection`)
- [x] Sanity: `first: 50` still returns exactly 50 rows
- [x] Root-level rejection: oversized inline / variable `first` returns 400 `BAD_USER_INPUT`
- [x] Root-level rejection: oversized `last`, `offset` on `entitiesConnection`
- [x] **Nested rejection (new):** oversized `first` on `relationsByFromEntityIdList` inside `entitiesConnection.nodes` returns 400 `BAD_USER_INPUT`
- [x] **Nested rejection (new):** oversized `last` on the same nested field returns 400 `BAD_USER_INPUT`
- [x] All existing tests pass
- [x] `tsc --noEmit` clean

### DIFF
--- a/api/src/kg/__tests__/paginationCapPlugin.test.ts
+++ b/api/src/kg/__tests__/paginationCapPlugin.test.ts
@@ -168,7 +168,7 @@ describe("PaginationCapPlugin", () => {
 	})
 
 	it("rejects oversized first on nested sub-collections", async () => {
-		// Nested collections inside an entity selection (e.g. relationsByFromEntityIdList
+		// Nested collections inside an entity selection (e.g. relationsList
 		// or valuesList) must be capped too. The previous implementation used
 		// makeWrapResolversPlugin, which per PostGraphile docs only reliably
 		// influences SQL for root-level resolvers. The current implementation
@@ -179,7 +179,7 @@ describe("PaginationCapPlugin", () => {
 				entitiesConnection(first: 1) {
 					nodes {
 						id
-						relationsByFromEntityIdList(first: 10000) { id }
+						relationsList(first: 10000) { id }
 					}
 				}
 			}
@@ -195,7 +195,7 @@ describe("PaginationCapPlugin", () => {
 				entitiesConnection(first: 1) {
 					nodes {
 						id
-						relationsByFromEntityIdList(last: 10000) { id }
+						relationsList(last: 10000) { id }
 					}
 				}
 			}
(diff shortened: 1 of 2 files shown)


api/src/kg/__tests__/paginationCapPlugin.test.ts
```diff
@@ -60,4 +91,409 @@ describe("PaginationCapPlugin", () => {
...
+			{
+				entities { id }
+			}
+		`)
+
+		expect(result.status).toBe(200)
+		expect(result.body.errors).toBeUndefined()
+	})
+
+	it("rejects oversized first on nested sub-collections", async () => {
```
0xJagger, line 170: 💎

0xJagger (approved):
Beautiful tests 🤝

0xyaco (approved):

PR #592, opened 2026-04-17
feat: log origin on requests

0xyaco:
## Summary
Add the `Origin` request header as an OTEL span attribute (`http.request.header.origin`) in `canonicalRequestLogging`. The existing `SentrySpanProcessor` forwards span attributes onto Sentry transactions, so every traced request now carries the caller's origin — making it trivial to filter or group transactions by frontend (production app, testnet, localhost, etc.) and to spot unexpected clients hitting the API.

Attribute is only set when the header is present, so server-to-server / curl callers don't leave an empty tag.

## Why
Sentry transactions currently don't distinguish between different frontends calling the same endpoint — during an error spike we have to correlate by path/IP, which loses the signal once multiple apps share the API. `Origin` is a CORS-standard header browsers always send, and Sentry indexes span attributes as searchable tags, so this is the cheapest path to per-frontend visibility.

## Out of scope
- No change to the canonical `log.info(...)` start/end lines — origin goes on the span only, to keep stdout noise unchanged and let Sentry be the query surface.
- No PII scrubbing logic — `Origin` is the scheme+host (never a path or query), so it's safe as-is.

## Test plan
- [ ] With `SENTRY_DSN` + `SENTRY_TRACES_SAMPLE_RATE=1.0` set locally, hit `/graphql` with `-H 'Origin: https://example.com'` and confirm the resulting transaction in Sentry shows `http.request.header.origin: https://example.com` under Tags/Contexts
- [x] Same request without an `Origin` header produces a transaction with no `http.request.header.origin` attribute (verified locally via `ConsoleSpanExporter`)
- [x] Existing span attributes (`http.method`, `http.route`, `http.status_code`, `http.request_id`) still appear unchanged


0xJagger (approved):

0xNeoDev:
for the future ai's, `X-Real-IP` `X-Forwarded-For` unspoofable headers for real client ip

PR #642, opened 2026-04-27
fix(kg-indexer): allow root space to publish system property schema

0xJagger:
## Summary

`filter_protected_relations` was dropping every edit op whose `from`/`to` is one of the four onchain-derived system property entities (`SpaceAddress`, `VotingMode`, `SpaceId`, `CreatedAtBlock`). That over-broadly drops the legitimate seed edit defining those entities' schema (e.g. `SpaceAddress -types-> Property`, `SpaceAddress -DataType-> Text`), so the system properties never get their type/data-type relations in the graph.

This PR adds a `ROOT_SPACE_ID` constant in the SDK and gates the `from`/`to` check on the publishing space — edits from root pass through, all others still drop. The relation-type filter (`SYSTEM_TYPES_RELATION_TYPE_ID`) and the value filter remain unconditional even from root: those are owned by `map_space_registered` from onchain events, and allowing edit-pipeline writes would race two writers on the same rows.

`space_id` cannot be spoofed — `SpaceRegistry.enter()` enforces it via `msg.sender` lookup or signature verification, so gating on it is sound defense.

A new ADR-003 in `kg-indexer/docs/DECISIONS.md` records the protection model.

## Test plan

- [x] `cargo build -p kg-indexer` clean
- [x] `SQLX_OFFLINE=true cargo test -p kg-indexer --bins handlers::edits` — 36/36 pass, including 5 new tests covering root-passes, root-still-blocked-from-system-types, root-still-blocked-from-protected-value, and a seed-edit smoke test
- [ ] Post-deploy: confirm the queued root-space relation lands the missing schema relations on the four system property entities

### DIFF
--- a/kg-indexer/src/handlers/edits.rs
+++ b/kg-indexer/src/handlers/edits.rs
@@ -1483,62 +1483,37 @@ mod tests {
 
     #[test]
     fn seed_edit_smoke_root_space_passes_property_schema() {
-        // Regression test for the original bug: the seed edit defining the
-        // schema for the four protected property entities had its 8
-        // schema-defining relations silently filtered out.
-        //
-        // Construct 8 CreateRelation ops with the same shape as the seed
-        // edit — for each protected property entity, one
-        // `<property> -types-> Property` relation and one
-        // `<property> -DataType-> Text` relation — and assert all 8 survive
-        // when published from the root space, and that all 8 are dropped when
-        // published from any other space.
-        let space_address = Uuid::parse_str(sdk::core::ids::SPACE_ADDRESS_PROPERTY_ID).unwrap();
-        let voting_mode = Uuid::parse_str(sdk::core::ids::VOTING_MODE_PROPERTY_ID).unwrap();
-        let space_id_prop = Uuid::parse_str(sdk::core::ids::SPACE_ID_PROPERTY_ID).unwrap();
-        let created_at_block =
-            Uuid::parse_str(sdk::core::ids::CREATED_AT_BLOCK_PROPERTY_ID).unwrap();
-
-        // `type` relation type — exists in the SDK.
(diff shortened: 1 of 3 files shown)


kg-indexer/src/handlers/edits.rs
```diff
@@ -110,21 +113,35 @@ fn extract_context(ctx: &Option<Context>) -> (Option<Uuid>, Option<Uuid>) {
...
+fn filter_protected_relations(ops: Vec<RelationOp>, edit_space_id: &Uuid) -> Vec<RelationOp> {
+    let is_root = edit_space_id == &*ROOT_SPACE_UUID;
     ops.into_iter()
         .filter(|op| match op {
             RelationOp::Create(rel) => {
                 let type_protected = PROTECTED_RELATION_TYPES.contains(&rel.type_id);
                 let from_protected = PROTECTED_PROPERTIES.contains(&rel.from_id);
                 let to_protected = PROTECTED_PROPERTIES.contains(&rel.to_id);
-                !type_protected && !from_protected && !to_protected
+                !type_protected && (is_root || (!from_protected && !to_protected))
```
0xNeoDev, line 144: so only relations `to` a system entity are allowed now? should `from` be allowed too? @0xJagger

0xJagger: Yes, both relations from/to system entities created by the root space should be allowed

This is the truth table on the return statement

| type_protected | is_root | from/to_protected | result |
|---|---|---|---|
| true | * | * | **drop** (type rule wins) |
| false | true | true | **keep** (root override) |
| false | true | false | keep |
| false | false | true | **drop** |
| false | false | false | keep |

So if `is_root` then we accept any relation from/to system properties

0xNeoDev (approved):

PR #662, opened 2026-04-30
feat(monitoring): unify Environment dropdown across api/knowledge/search/scoring + ingress filters

0xNeoDev:
Replaces the existing `$namespace` template variable on the Gaia Overview dashboard with a single `$env` suffix variable that drives every env-discriminating filter: namespace filters across all four service namespaces, the api-ingress filter, and the per-env values exposed to HPA via prometheus-adapter. Picking "Staging" in the dropdown switches api 5xx counts, RPS, GraphQL pool utilization, GraphQL histograms, HPA replicas, and per-service mem/cpu panels in one go.

### `api-ingress-rules.yaml` — recording rules

Was: `host="testnet-api.geobrowser.io"` hardcoded; `host` label aggregated away.

Now: `exported_namespace=~"api(-staging)?",ingress="api",exported_service="api"` with `sum by (exported_namespace, ...)` to preserve the env discriminator. Namespace-based filters survive DNS migrations; the `ingress`/`exported_service` triple rejects unrelated ingresses (e.g. monitoring/grafana). Cluster-wide aggregates still possible via `sum(...)` wrapping.

### `gaia-overview-dashboard.yaml` — main dashboard

- `$env` template variable (custom, label "Environment", text="Production" value="(?:)" / text="Staging" value="-staging"). The `(?:)` value works around Grafana's empty-value substitution bug — it's a regex-empty group that's a non-empty string, so Grafana substitutes it cleanly and the regex engine treats it as nothing.
- All 18 PromQL `expr` strings + 14 URL-encoded data-link drill-ins rewritten to use `${env:raw}` (the `:raw` formatter bypasses Grafana's regex auto-escape so `(?:)` reaches the regex engine intact).
- All filters now anchored regex matches: `namespace=~"^api${env:raw}$"`, `exported_namespace=~"^api${env:raw}$"`, etc. Anchors prevent substring matches between `api` and `api-staging`.
- `gaia_api_*` panels (GQL Pool Saturated, REST/GraphQL Pool Utilization, Pool Waiting & Acquire Timeouts, GraphQL Query Cost histograms, GraphQL Response Size histograms) now filter by `namespace=~"^api${env:raw}$"` — picking Staging routes them to `api-staging`.
- Multi-namespace regex panels (Pod Restarts, Deployment Availability, CPU/Memory by Service): `namespace=~"^(api|knowledge|search|scoring)${env:raw}$"`.

### `api-ingress-dashboard.yaml` — ingress dashboard

- Same `$env` variable added.
- All 14 recording-rule references and the one direct `nginx_ingress_controller_requests` query rewritten to filter by `exported_namespace=~"^api${env:raw}$",ingress="api",exported_service="api"`.

### `api-capacity-alerts.yaml` — PrometheusRule alerts

- `Api5xxRateHigh` and `ApiHpaMaxedWithHighP99` exprs pinned to the prod env via `exported_namespace="api",ingress="api",exported_service="api"` to preserve pre-existing prod-only alert behavior. Without this, the recording-rule label change would cause per-host alerts and break the vector matching on the HPA `and`.

### `prometheus-adapter-config.yaml` — HPA external metrics

- `api_ingress_p95_latency_seconds_2m` and `api_ingress_p99_latency_seconds_2m` external metrics pinned to prod env via the same triple. The 503 ratio rule and the raw `nginx_ingress_controller_requests` `seriesQuery` were also updated to use namespace-based filtering for consistency.

After applying, the prometheus-adapter pods need a restart to pick up the new ConfigMap (the adapter only reads its config at startup):

```bash
kubectl -n monitoring rollout restart deployment prometheus-adapter
```

## What stays cluster-wide (intentional)

OpenSearch panels (Search QPS, JVM Heap, Cluster Health, etc) — single shared OpenSearch cluster, so per-env doesn't apply.

## Verification

- `kubectl apply --dry-run=client` clean on all 5 files.
- All hardcoded `host="testnet-api.geobrowser.io"` references removed from the repo.
- Dashboard JSON parses; templating + panel counts intact (65 + 9).
- Probed live cluster: `exported_namespace` label values present (`api`, `api-staging`); ingress filter matches both; recording rule emits 2 series (one per env); prod p99 = 0.44s, staging = 0 (no traffic).

## Deploying

```bash
kubectl apply -f monitoring/k8s/api-ingress-rules.yaml \
              -f monitoring/k8s/api-ingress-dashboard.yaml \
              -f monitoring/k8s/gaia-overview-dashboard.yaml \
              -f monitoring/k8s/api-capacity-alerts.yaml \
              -f monitoring/k8s/prometheus-adapter-config.yaml
kubectl -n monitoring rollout restart deployment prometheus-adapter
```

Grafana sidecar reloads the dashboard ConfigMaps within ~10s. Recording rules and alerts apply immediately. The adapter rollout takes ~10s.

### DIFF
--- a/monitoring/k8s/api-capacity-alerts.yaml
+++ b/monitoring/k8s/api-capacity-alerts.yaml
@@ -29,7 +29,7 @@ spec:
             kube_horizontalpodautoscaler_status_current_replicas{namespace="api", horizontalpodautoscaler="api"}
             >=
             kube_horizontalpodautoscaler_spec_max_replicas{namespace="api", horizontalpodautoscaler="api"}
-            and api:ingress_p99_latency_seconds:5m > 2
+            and api:ingress_p99_latency_seconds:5m{exported_namespace="api",ingress="api",exported_service="api"} > 2
           for: 10m
           labels:
             severity: critical
@@ -39,7 +39,7 @@ spec:
             description: API is at max replicas with sustained p99 latency above 2s.
 
         - alert: Api5xxRateHigh
-          expr: api:ingress_5xx_ratio:rate5m > 0.02
+          expr: api:ingress_5xx_ratio:rate5m{exported_namespace="api",ingress="api",exported_service="api"} > 0.02
           for: 10m
           labels:
             severity: warning
--- a/monitoring/k8s/api-ingress-dashboard.yaml
+++ b/monitoring/k8s/api-ingress-dashboard.yaml
@@ -31,6 +31,30 @@ data:
               "text": "Prometheus",
               "value": "Prometheus"
             }
+          },
+          {
+            "name": "env",
+            "label": "Environment",
+            "description": "Suffix appended to the ingress host. Empty for production, '-staging' for staging.",
+            "type": "custom",
+            "query": "Production : (?:),Staging : -staging",
+            "current": {
+              "selected": true,
+              "text": "Production",
+              "value": "(?:)"
+            },
+            "options": [
+              {
+                "selected": true,
+                "text": "Production",
+                "value": "(?:)"
+              },
+              {
+                "selected": false,
+                "text": "Staging",
+                "value": "-staging"
+              }
+            ]
           }
         ]
       },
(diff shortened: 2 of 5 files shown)


ohohoreilly:
Closing as obsolete — this predates the v2 namespace migration.

The filters here key on namespaces `api` and `api(-staging)?`. Neither exists in the cluster any more; the live stack runs in namespace `gaia`. Porting it as-is would give dashboards that filter on namespaces with nothing in them, so the unified $env dropdown would switch between two empty views.

The idea is still good and the work was sound when written — it just needs redoing against the current namespace layout rather than rebasing.

Worth flagging separately: the monitoring config already on `staging` has the same problem — `monitoring/k8s/*.yaml` also filters on `api` and `api-staging`, so those dashboards may have been blind since the v2 migration. That's a bigger question than this PR.

PR #713, opened 2026-06-05
test(scoring): cover RankingConfig anti-sybil mutual-exclusion invariant

ohohoreilly:
## What

Adds test coverage for `RankingConfig.__post_init__` in the scoring cronjob.

`use_distance_weighting` and `filter_non_members` are two alternative anti-sybil mechanisms, and the config treats them as mutually exclusive — enabling both raises a `ValueError` at construction time. This invariant was previously untested.

## Tests added (`scoring-service/cronjob/tests/test_ranking_config.py`)

- Enabling both mechanisms raises `ValueError`
- Distance weighting alone is allowed
- Member filtering alone is allowed
- Neither enabled is allowed

## Notes

- Additive only — no production code changed.
- Full cronjob suite green: 88 passed.
- Run with `uv run --extra dev pytest`.

### DIFF
--- a/scoring-service/cronjob/tests/test_ranking_config.py
+++ b/scoring-service/cronjob/tests/test_ranking_config.py
@@ -1,0 +1,46 @@
+"""Validation tests for RankingConfig.__post_init__.
+
+Distance weighting and member filtering are two different anti-sybil
+mechanisms, and the config treats them as mutually exclusive: enabling both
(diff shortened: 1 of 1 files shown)


ohohoreilly:
Superseded by #715 (re-opened as a same-repo PR now that I have write access; identical change).

PR #717, opened 2026-06-05
test(scoring): characterize personal-space vote filtering

ohohoreilly:
## What

Characterization tests for `EntityScorer.filter_valid_votes`, focused on the personal-space edge case.

## Behavior pinned

A personal space has no rows in `members`/`editors`, so its owner never appears in any voter's `member_spaces`/`editor_spaces`. With `filter_non_members=True`, `filter_valid_votes` therefore drops **every** vote cast in a personal space — **including the owner's own vote**.

This is a known gap (the owner arguably should count for their own personal space). These tests document the behavior as-is, so a future *owner-counts* fix has a guard that flips from "documents the gap" to "protects the fix."

## Tests added (`scoring-service/cronjob/tests/test_filter_valid_votes.py`)

- Owner's vote in their personal space is **dropped** when `filter_non_members=True`
- Same vote is **kept** when filtering is disabled (proves the vote is otherwise valid)
- Vote is **kept** when a membership row exists (positive control — isolates missing membership as the cause)

## Notes

- Additive only — no production code changed.
- Full cronjob suite green: 87 passed.
- Run with `uv run --extra dev pytest`.

### DIFF
--- a/scoring-service/cronjob/tests/test_filter_valid_votes.py
+++ b/scoring-service/cronjob/tests/test_filter_valid_votes.py
@@ -1,0 +1,103 @@
+"""Characterization tests for EntityScorer.filter_valid_votes.
+
+These pin the current behavior of member filtering, with focus on the
+personal-space edge case.
+
+A personal space has no rows in the `members`/`editors` tables, so the
+space owner is not present in any user's `member_spaces`/`editor_spaces`.
+When `filter_non_members=True`, `filter_valid_votes` therefore drops *every*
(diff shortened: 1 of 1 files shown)


0xJagger (approved):
Nice catch! We should probably rethink the members/editors filter, since personal spaces don’t follow the same logic as DAO spaces:
- For DAO spaces, we can use the members/editors filter
- For personal spaces, we should either count all votes or define a new anti-Sybil method

PR #719, opened 2026-06-05
perf(scoring): index users by id in the entity ranking hot loop

ohohoreilly:
## What

`rank_entities` calls `EntityScorer.filter_valid_votes` **once per entity**, and that method located each vote's user with a linear scan over all users:

```python
user = next((u for u in users if u.id == vote.user_id), None)
```

So entity ranking was **O(entities × votes × users)**. `apply_distance_weighting` had the same per-vote scan (**O(votes × users)**).

## Change

Build a `{user_id: user}` dict once per call and look up via `.get()`:

- `filter_valid_votes` and `apply_distance_weighting` drop to **O(votes + users)**.
- A small `_index_users_by_id` helper keeps the **first** occurrence of any duplicate id, exactly matching the previous `next()` behavior — so this is a pure speedup with no behavior change.

## Verification

- Full cronjob suite green: 84 passed (`uv run --extra dev pytest`).
- The member-filtering keep/drop paths are additionally characterized in #717.

`+16 / -2`, single file.

### DIFF
--- a/scoring-service/cronjob/src/algorithm/scoring.py
+++ b/scoring-service/cronjob/src/algorithm/scoring.py
@@ -68,10 +68,23 @@ class EntityScorer:
 
         return hot_score
 
+    @staticmethod
+    def _index_users_by_id(users: list[User]) -> dict[str, User]:
+        """Index users by id for O(1) lookup.
+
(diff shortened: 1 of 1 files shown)


0xJagger (approved):
lgtm!

PR #720, opened 2026-06-05
feat(api): make entities_ordered_by_property filter-first and index-ordered

0xJagger:
## Before and after

Demo loading times on https://www.geobrowser.io/space/19f11bc6f1a62ac434936af814d1f8b5?tabId=77de05b07bac47f78b313b36c293d6de

**Before** — takes ~30s to load one page of info

https://github.com/user-attachments/assets/1435a2ef-a838-49b5-ae65-38ca11247c18

**After** — local geobrowser using the new function, load time <1s

https://github.com/user-attachments/assets/4bd946e2-81de-4965-91cf-08daabd5944e

## What this PR does

Rewrites `entities_ordered_by_property` to be **filter-first**. The old version materialized every entity with the sort property in the space, sorted the whole set with a non-indexable `ORDER BY CASE …`, and only then let PostGraphile apply the type filter + page limit — so cost scaled with the entire property set rather than the page. This version drives from the selective space/type set first, using **existing** `relations`/`values` indexes.

It's a **function-only migration** — just `CREATE OR REPLACE FUNCTION` (+ dropping the old signatures). No table, column, or index changes.

- **Conditional SQL** — the WHERE clause emits a predicate only for each non-null param, with **no `param IS NULL OR …` guards**. That guard was the key blocker: with a bound parameter the planner can't collapse it into a semi-join, so it scanned the full property set and ran the type `EXISTS` per row. Building the clause conditionally lets the planner use indexes and turn the type filter into a semi-join that drives from the selective type set.
- **New optional `type_ids uuid[]` param** — pushes the type filter inside the query (`to_entity_id = ANY(...)`, OR semantics, scoped to the value row's space). It only needs to be a *necessary superset*; the app still applies the exact filter as a PostGraphile residual. Backward compatible: omitting it reproduces the prior (no type filter) behavior, so the backend can ship ahead of the frontend.

## Correctness

- **One row per entity** — `DISTINCT ON (entity_id)`, since `values` has no uniqueness constraint on `(entity_id, property_id, space_id)` (verified unique in the data, but not enforced).
- **Stable pagination** — tie-break by `e.id`, so low-cardinality sorts (e.g. boolean) don't reorder between pages.
- **Space-scoped type filter** — the type `EXISTS` constrains `r.space_id = v.space_id`, so a type assignment in another space can't leak in.
- **Graceful on bad input** — unsupported/unresolved `data_type` returns an empty set instead of raising; NULL `property_id` is handled safely (`quote_nullable`) instead of crashing at `EXECUTE`.
- **Deterministic data-type resolution** — the `DATA_TYPE` lookup is `ORDER BY r.id LIMIT 1`.

## Why no new indexes

An earlier revision added 9 per-type composite sort indexes. Benchmarking on a testnet-scale dump showed they **aren't used** by any path the app issues: type-filtered sorts (the common case) are driven by the existing `relations`/`values` indexes, and space-only sorts do a bounded sort. A dedicated typed-column index only pays off if the page `LIMIT` is pushed *inside* the function — and as a plpgsql `SETOF` function the outer `LIMIT` can't reach the inner query. So the indexes would be pure write amplification on `values` (the hottest table). They're dropped here; the index-ordered + limit-early path is a follow-up (see below).

## Measurements (testnet-scale local dump: 27M entities / 3M values / 2.5M relations, warm)

| Path | Before | After |
|---|--:|--:|
| **Type-filtered** (a space's table scoped to a type — the common case) | ~30s | **~15–20ms** SQL |
| Space-only (no type filter) | several s | **~1s** |

Results verified **identical** to the previous function (same rows/order on the reference query), the duplicate-row artifact from the old stacked `name`/`spaceIds` filter is gone, and multi-type OR/AND + type-with-value filters were validated end-to-end. (Warm numbers; the first query after a cold start is slower.)

## Migration safety

Function-only — no index builds, no table rewrite, no locks. `CREATE OR REPLACE FUNCTION` + `DROP FUNCTION IF EXISTS` of the prior signatures. Nothing to coordinate at deploy.

## Follow-ups (not in this PR)

- **Index-ordered + limit-early** — to make space-only sorts sub-millisecond, push the page bound into the function (keyset or offset pagination) and reintroduce the per-type indexes (now actually used). Frontend-coupled (the query builder passes the bound in), so it's its own change.
- **Frontend query builder** — pass the necessary type superset to the new `typeIds` arg and keep the exact filter as a residual. Documented for the frontend team; backward compatible with this change.
- Regression test for the function (dedup / stable order / multi-type / empty-on-bad-input); confirm the `Float64`-style data-type-name fallback.

## Test plan

- [x] Migration applies cleanly (`drizzle-kit`-generated journal/snapshot; SQL validated against the dump)
- [x] Function returns identical rows/order vs the previous definition (incl. on a glibc DB matching prod collation)
- [x] Single-type, multi-type OR, multi-type AND, and type + value validated end-to-end via GraphQL
- [x] Edge cases: unsupported `data_type` → empty; NULL `property_id` → empty (no crash); stable order across repeated runs

### DIFF
--- a/api/drizzle/0060_optimize-entities-ordered-by-property.sql
+++ b/api/drizzle/0060_optimize-entities-ordered-by-property.sql
@@ -1,0 +1,94 @@
+-- GEO-675: optimize entities_ordered_by_property (sorted table/collection views).
+--
+-- Orders entities by one of a property's typed values, filter-first: it drives from the
+-- selective space/type set via existing relations/values indexes and a type semi-join,
+-- instead of the old version which sorted the whole property set with a non-indexable
+-- ORDER BY CASE before filtering. The sort column is resolved from `data_type`, or the
+-- property's DATA_TYPE relation (ORDER BY r.id for determinism; empty set on an
+-- unsupported/unresolved type). Predicates are appended only for the args present (no
+-- "arg IS NULL OR ..." guards) so the planner can use indexes and the type semi-join.
+-- `type_ids` is an optional, space-scoped NECESSARY type superset (OR semantics) — the app
+-- still applies the exact filter as a PostGraphile residual, and omitting it reproduces the
+-- prior no-type behavior (backward compatible). DISTINCT ON returns one row per entity; the
+-- e.id tie-break keeps pagination stable.
+--
+-- No dedicated sort indexes
+-- Type-filtered sorts (the common path) are already index-driven via the existing
+-- relations/values indexes; space-only sorts fall back to a bounded sort.
+
+DROP FUNCTION IF EXISTS public.entities_ordered_by_property(uuid, uuid, sort_order, text);
+--> statement-breakpoint
+DROP FUNCTION IF EXISTS public.entities_ordered_by_property(uuid, uuid, sort_order, text, uuid);
+--> statement-breakpoint
+CREATE OR REPLACE FUNCTION public.entities_ordered_by_property(
+  property_id uuid,
+  space_id uuid DEFAULT NULL,
+  sort_direction sort_order DEFAULT 'ASC',
+  data_type text DEFAULT NULL,
+  type_ids uuid[] DEFAULT NULL
+)
+RETURNS SETOF entities AS $fn$
+DECLARE
+  resolved_type text;
+  sort_expr     text;
+  null_pred     text;
+  dir           text;
+  space_pred    text := '';
+  type_pred     text := '';
+  sql           text;
+BEGIN
+  IF data_type IS NOT NULL THEN
+    resolved_type := lower(data_type);
+  ELSE
+    SELECT lower(v."text")
+      INTO resolved_type
+      FROM relations r
+      JOIN "values" v ON v.entity_id = r.to_entity_id
+     WHERE r.from_entity_id = entities_ordered_by_property.property_id
+       AND r.type_id    = '6d29d578-49bb-4959-baf7-2cc696b1671a'
+       AND v.property_id = 'a126ca53-0c8e-48d5-b888-82c734c38935'
+     ORDER BY r.id
+     LIMIT 1;
+  END IF;
+
+  CASE resolved_type
+    WHEN 'text'     THEN sort_expr := 'left(v."text", 1024)'; null_pred := 'v."text" IS NOT NULL AND length(btrim(v."text")) > 0';
+    WHEN 'integer'  THEN sort_expr := 'v.integer';            null_pred := 'v.integer IS NOT NULL';
+    WHEN 'float'    THEN sort_expr := 'v.float';              null_pred := 'v.float IS NOT NULL';
+    WHEN 'decimal'  THEN sort_expr := 'v."decimal"';          null_pred := 'v."decimal" IS NOT NULL';
+    WHEN 'boolean'  THEN sort_expr := 'v.boolean';            null_pred := 'v.boolean IS NOT NULL';
+    WHEN 'date'     THEN sort_expr := 'v.date';               null_pred := 'v.date IS NOT NULL AND length(btrim(v.date)) > 0';
+    WHEN 'time'     THEN sort_expr := 'v."time"';             null_pred := 'v."time" IS NOT NULL AND length(btrim(v."time")) > 0';
+    WHEN 'datetime' THEN sort_expr := 'v.datetime';           null_pred := 'v.datetime IS NOT NULL AND length(btrim(v.datetime)) > 0';
+    WHEN 'point'    THEN sort_expr := 'v.point';              null_pred := 'v.point IS NOT NULL AND length(btrim(v.point)) > 0';
+    ELSE
+      RETURN;
+  END CASE;
+
+  dir := CASE WHEN sort_direction::text = 'DESC' THEN 'DESC' ELSE 'ASC' END;
+
+  IF space_id IS NOT NULL THEN
+    space_pred := format(' AND v.space_id = %L', space_id);
+  END IF;
+
+  IF type_ids IS NOT NULL AND cardinality(type_ids) > 0 THEN
+    type_pred := format(
+      ' AND EXISTS (SELECT 1 FROM relations r WHERE r.from_entity_id = v.entity_id AND r.type_id = %L AND r.to_entity_id = ANY(%L::uuid[]) AND r.space_id = v.space_id)',
+      '8f151ba4-de20-4e3c-9cb4-99ddf96f48f1', type_ids);
+  END IF;
+
+  -- OPTIMIZE (follow-up): LIMIT/OFFSET are applied by PostGraphile outside this function, so
+  -- PG materializes the full ordered set before paginating. Pushing them inside would be
+  -- limit-early but breaks cursor pagination; deferred since the type-prefiltered set is small.
+  sql :=
+       'SELECT e.* FROM ('
+    || '  SELECT DISTINCT ON (v.entity_id) v.entity_id, ' || sort_expr || ' AS sort_value'
+    || '  FROM "values" v'
+    || '  WHERE v.property_id = ' || quote_nullable(property_id) || ' AND ' || null_pred || space_pred || type_pred
+    || '  ORDER BY v.entity_id, ' || sort_expr || ' ' || dir
+    || ') sub JOIN entities e ON e.id = sub.entity_id'
+    || ' ORDER BY sub.sort_value ' || dir || ', e.id';
+
+  RETURN QUERY EXECUTE sql;
+END;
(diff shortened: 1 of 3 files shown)


ohohoreilly:
## Review

Nice work — pulled this onto a local Postgres with your migration applied and a 500k-row synthetic set (one integer property across 40 spaces) and ran `EXPLAIN ANALYZE`. **The filter-first rewrite is a real, big win for the space/type-scoped paths the app actually uses**, and the correctness hardening (DISTINCT ON, `e.id` tiebreak, space-scoped type EXISTS, graceful empty-set, deterministic type resolution) is genuinely thorough. Injection surface looks clean too (`data_type` only reaches SQL via the whitelist `CASE`; ids via `format %L` / `quote_nullable`).

A few things the measurements surfaced that are worth resolving before merge.

### Measured (500k rows, `LIMIT 100`, local synthetic — trust ratios/shapes, not absolute ms)

| Path | End-to-end |
|---|--:|
| Filtered (single space) | **139 ms** |
| Unfiltered (no space, no type) | **1446 ms** |

### 1. The 9 new indexes aren't actually used by the function as written — `DISTINCT ON` blocks them

In the *filtered* plan, the leaf access is the **existing** `values_space_id_idx` (bitmap), followed by a `Sort` — none of the new `values_sort_*` indexes appear:

```
->  Sort  (Sort Key: v.entity_id, v."integer" DESC)        -- forced by DISTINCT ON (entity_id)
      ->  Bitmap Heap Scan on "values" v
            ->  Bitmap Index Scan on values_space_id_idx     -- not values_sort_integer_idx
```

The reason: `DISTINCT ON (v.entity_id)` requires the inner `ORDER BY` to lead with `entity_id`, but the indexes are `(property_id, space_id, <col>)`. So the planner can't use them for ordering and falls back to a sort. The 139 ms win comes entirely from **filter pruning** (space → ~12.8k rows → sort), not from index-ordering.

Control query — same filter, **without** `DISTINCT ON`, `ORDER BY integer LIMIT 100`:

```
Limit  ->  Index Scan Backward using values_sort_integer_idx
Execution Time: 0.6 ms
```

So the indexes *work* and would give a ~0.6 ms index-ordered top-N — `DISTINCT ON` is what defeats them (and also blocks `LIMIT` push-down).

**Implication:** as written, the 9 partial composite indexes add write amplification to `values` (the hottest table — kg-indexer + the score mirror write it constantly) for ~no read benefit. Two ways to reconcile:
- **(a)** Restructure so the index can drive the order — e.g. drop `DISTINCT ON` where `(entity_id, property_id, space_id)` is effectively unique (worth confirming that invariant), or dedup a different way, so it becomes the 0.6 ms `Index Scan + LIMIT` path. This would also resolve the deferred LIMIT-early follow-up.
- **(b)** If we're keeping `DISTINCT ON`, drop most of the 9 indexes since the planner doesn't use them — the existing `values_property_space_idx` already serves the filter.

### 2. The unfiltered path is unimproved (~1.4 s) and the same shape as the old function

With no `space_id`/`type_ids`, the function materializes + sorts the whole property set (external merge sort, spilled to disk for 500k rows) and the outer `LIMIT` can't push into the `plpgsql` function. That's essentially identical to the pre-rewrite cost. Fine *if* the app always passes a space or type — but there's no guard. Is there any call path (e.g. a global sort on a high-cardinality property like `name`) that would hit this? If so, worth a guard or the (a) restructure above.

### 3. Deploy footgun (operational)

The 9 `CREATE INDEX IF NOT EXISTS` build non-concurrently inside the migration transaction unless pre-created with `CONCURRENTLY`. On `values` (~3M rows, constantly written) a non-concurrent build takes an `ACCESS EXCLUSIVE` lock and stalls ingestion for the duration. The manual pre-create plan is correct, but it relies on remembering — recommend putting it in the runbook / gating it in the deploy, not just the PR body. (Lower stakes if #1(b) trims the index count.)

### 4. No automated test

This is a perf- and correctness-sensitive function (dynamic SQL, dedup, stable order, multi-type OR/AND, empty-on-bad-input). Worth at least one regression test for those cases — there's a skipped `entitiesOrderedByScore` test to model from.

### 5. Confirm fallback type-name resolution (not a regression)

The relation-resolution fallback matches `lower(v."text")` against `'integer'`/`'float'`/etc. The native data-type entities are named e.g. `Float64`/`Datetime` in places — if a property's `DATA_TYPE` name is `"Float64"`, `lower()` → `float64` misses the `'float'` branch and silently returns empty. Same labels as the old function, so not a regression, but worth confirming the fallback path matches (the explicit `data_type` arg path is fine).

### Minor
- 2nd `DROP FUNCTION … (uuid, uuid, sort_order, text, uuid)` is a scalar `uuid`, not `uuid[]` — harmless no-op, but doesn't match the new array signature.
- `quote_nullable(property_id)` vs `format('%L', space_id)` — both safe, just inconsistent.

---

Net: the filter-first direction is a clear win and worth shipping. The main thing to resolve is the **`DISTINCT ON` ↔ indexes tension** in #1 — either make the indexes earn their write cost (and unlock the 0.6 ms path) or drop them — plus the deploy safety in #3. Happy to share the full harness/EXPLAIN output if useful.

0xJagger:
Thanks for the incredibly thorough review — pulling it onto a synthetic 500k set and running `EXPLAIN ANALYZE` is above and beyond, and the `DISTINCT ON` ↔ index finding is a great catch. 🙏

**1. `DISTINCT ON` defeats the indexes** — you're right, and the irony is that the `DISTINCT ON (entity_id)` was added to satisfy a duplicate-row finding from the previous review round, and it's exactly what blocks the index-ordered path. We're going with your option (a): drop `DISTINCT ON` and restore the index-ordered top-N (the ~0.6 ms path), which also lets `LIMIT` push in. The catch is it hinges on `(entity_id, property_id, space_id)` being effectively unique — and you're right that the schema doesn't enforce it (PK is the value `id`, only a non-unique composite index), so we'll confirm it on real data before dropping the dedup. If it holds — expected, since multi-valued attributes are modeled as relations rather than multiple value rows — the indexes earn their write cost and the dedup is unnecessary. Will share re-benchmarks once it's in.

The remaining points are all good catches — capturing them as follow-ups rather than expanding this PR:

**2. Unfiltered path / no guard** — same root cause; the (a) restructure makes this an index-ordered top-N too. Worth confirming there's no global high-cardinality sort path; we'll track it.

**3. CONCURRENTLY deploy footgun** — agreed it shouldn't live only in the PR body; we'll move the manual pre-create into the deploy runbook / gate it (lower stakes if (a) lets us trim the index set).

**4. Regression test** — agreed, worth one for dedup / stable order / multi-type OR-AND / empty-on-bad-input, modeled on the skipped `entitiesOrderedByScore` test.

**5. Float64 fallback** — good eye; `lower("Float64") → 'float64'` would miss the `'float'` branch. Same as the old function and the app passes `dataType` explicitly so the fallback isn't hit in practice, but we'll confirm the data-type entity names and tighten it.

**Minors** — the scalar-`uuid` `DROP` is intentional (it drops a single-`uuid` prototype overload that only ever existed on local DBs), but I'll reconcile the `quote_nullable` / `format %L` inconsistency.

Net: doing the #1 restructure in this PR, the rest as follow-ups. Thanks again — really useful review. 🙏

0xJagger:
**Update on #1 (`DISTINCT ON` ↔ indexes):** after benchmarking on a testnet-scale dump, we landed on a different resolution than my earlier reply — we **kept `DISTINCT ON` and dropped the 9 indexes** (rather than dropping `DISTINCT ON`). Why:

- Dropping `DISTINCT ON` alone didn't unlock the index path: the plpgsql `SETOF` function is a black box, so the outer `LIMIT` can't reach the inner query. Without an inner `LIMIT` the planner still bitmap-scans + sorts the whole set — the `0.6ms` index path only appears once a `LIMIT` is *in* the query. It also destabilized the type-filtered plan (~1s).
- The indexes aren't used by any path the app issues anyway: type-filtered sorts (the common case) are driven by the existing `relations`/`values` indexes (~15–20ms), and space-only sorts do a bounded sort (~1s). So they'd be pure write amplification on the hot `values` table.

So this PR is now a **function-only migration** — filter-first rewrite, no index builds. The big win (type-filtered ~30s → ~20ms) is preserved via existing indexes; space-only is ~1s, which we're fine with for now. This also makes your #3 (CONCURRENTLY deploy footgun) moot — there are no index builds.

The proper **index-ordered + limit-early** path (your option (a), done with the `LIMIT` pushed into the function via keyset/offset pagination) is the way to get space-only sub-ms, but it's frontend-coupled, so we're tracking it as a follow-up.

PR #722, opened 2026-06-05
fix(notification-indexer): comment parent_id = immediate parent (cascade)

0xNeoDev:
## What

For general comment-thread notifications, the payload's `parent_id` was the **thread root** instead of the comment the reply actually replied to. This fixes it to the **immediate parent**, and adds tests verifying the behavior.

## Root cause

curator-app builds a reply's `Reply to` relations with a **cascade**: it carries forward the whole ancestor chain (root target first) and appends the immediate parent **last**. The indexer emits one `ProposalCommentInfo` per reply-to edge, all sharing one `comment_entity_id`-keyed idempotency key. The consumer processed them per-edge, so the **first-processed edge (the root) won** the payload's `parent_id`.

Recipients were always correct (thread participants ∪ root creator); only the payload's `parent_id` fidelity was off — it mattered only for an app wanting the precise "X replied to *your* comment" parent.

## Fix

- **storage** `find_deepest_reply_target(targets)` — among a comment's reply-to targets, the immediate parent is the one that replies to the most of the *others* (deepest in the cascade). Returns `None` for single-target / non-cascade comments so the caller falls back to the lone target.
- **consumer** — group reply-to edges per comment, **process each comment once** (this also removes the prior N× thread-resolution for deep comments), and set `parent_id` to the resolved immediate parent.

This is robust to op order (it uses the thread topology, not the edge order), and unchanged for single-reply-to comments (geogenesis proposal comments).

## Tests

- **Unit**: `extract_proposal_comments_cascade_reply_emits_edge_per_target_in_op_order` documents the cascade extraction (one info per edge, in op order, sharing one idempotency key).
- **E2E**: the Phase-2b reply now uses a cascade builder (`reply-to [root, parent]`) and asserts `parent_id` == the **immediate parent**, not the root — this fails without the fix.
- 83 unit + 200 e2e green.

## Background

Came out of cross-checking the indexer against curator-app's `create-comment.ts`. Detection/routing/recipients/author-exclusion all already matched curator's comment shape; this `parent_id` fidelity was the one gap.

### DIFF
--- a/notification-service/e2e-tests/src/main.rs
+++ b/notification-service/e2e-tests/src/main.rs
@@ -899,12 +899,14 @@ async fn produce_test_events(kafka_broker: &str) -> Result<(), Box<dyn std::erro
         .await
         .map_err(|(e, _)| e)?;
 
-    // Phase 2b: a reply into the seeded thread (reply → EXISTING_COMMENT). The
-    // indexer walks to the root, notifying the prior participant + the root's
-    // creator (home space), but NOT the reply's own author.
-    let thread_reply = make_comment_hermes_edit(
+    // Phase 2b: a reply into the seeded thread, using the curator-app CASCADE —
+    // reply-to the root FIRST (carried forward) and the immediate parent LAST.
+    // The indexer walks to the root (notifying the prior participant + the root's
+    // creator, but NOT the reply's author) and resolves the immediate parent
+    // (EXISTING_COMMENT) for the payload's parent_id rather than the root.
+    let thread_reply = make_cascade_comment_hermes_edit(
         REPLY_COMMENT_BYTES,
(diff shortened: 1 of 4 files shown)


ohohoreilly:
Shipped as #959 — your commit, cherry-picked onto `staging` with authorship intact, merged as `1ada3529`.

Nothing was wrong with this fix. It targeted `main`, which is 147 commits behind `staging` and hasn't moved since 2026-07-27, and gaia ships from `staging` — so it sat MERGEABLE-but-BLOCKED against a branch nobody merges. The bug was still live on `staging` today (`parent_id: info.proposal_id` at main.rs:999), which is why it was worth rescuing rather than closing.

Closing as superseded.

PR #724, opened 2026-06-09
feat(sdk): add rank + ranking-block system IDs for the ranking-indexer

ohohoreilly:
## What

Defines the user-authored rank shapes in the Rust `sdk` crate (`core/ids`) so the upcoming **ranking-indexer** can detect them on the `knowledge.edits` stream — the foundational, dependency-light first piece of that work.

## IDs added

**Rank submissions** — canonical, mirrored from the *merged* grc-20 SDK (`core/ids/system`):
- `RANK_TYPE_ID`, `RANK_TYPE_PROPERTY_ID`, `RANK_VOTES_RELATION_TYPE_ID`, `VOTE_ORDINAL_VALUE_PROPERTY_ID`, `VOTE_WEIGHTED_VALUE_PROPERTY_ID`

**Ranking blocks** — ⚠️ **PROVISIONAL**, sourced from [`geo-sdk#89`](https://github.com/geobrowser/geo-sdk/pull/89) (`feat/ranks-create-update`), which is **not yet merged**:
- `RANKING_BLOCK_TYPE_ID`, `RANK_BLOCK_RELATION_TYPE_ID`, `RANK_START_DATE_PROPERTY_ID`, `RANK_END_DATE_PROPERTY_ID`, `RANK_AGGREGATION_RESTRICTION_PROPERTY_ID`, `RANK_RESTRICTION_MEMBERS_AND_EDITORS_ID`

> The block IDs ride on an unmerged SDK PR — they should be re-pinned to the released values once #89 lands. Called out in a code comment too.

## Deliberately *not* included: `RANK_POSITION`

The indexer-owned output relation type is system-minted (not an SDK shape) and needs a canonical assignment before it can join `PROTECTED_RELATION_TYPE_IDS`. The convention here is `UUIDv5(GEO_SYSTEM_NAMESPACE, "relation_type:RankPosition")` (matching how `SYSTEM_TYPES_RELATION_TYPE_ID` is derived) → `7f74e35d-14c0-53de-812e-28e2fb96ddba`, pending sign-off from the design owner on the exact name string. Tracked as a follow-up.

## Test

`cargo test -p sdk core::ids` → 4 passed. Additions are raw constants outside any tested collection, so existing `ids` tests are unaffected.

PR #725, opened 2026-06-09
feat(api): add private ranks schema for the ranking-indexer

ohohoreilly:
## What

Creates the **ranking-indexer's private working cache** as a dedicated `ranks` schema (migration `0061`). This is workstream #1 of the ranking-indexer build, and it's independent of the design's open questions (eligibility/scoring/publish semantics) — so it's safe to land now.

Four tables, per §6 of the ranking-indexer design:

| Table | Purpose | Key |
|---|---|---|
| `ranks.ranking_blocks` | one row per Ranking Block | `id` |
| `ranks.rankings` | one row per Rank submission | `id` |
| `ranks.ranking_items` | decoded items per submission | `(ranking_id, entity_id, space_id)` |
| `ranks.ranking_scores` | computed aggregate | `(block_id, entity_id, space_id)` |

## Why a private `ranks` schema

PostGraphile introspects only `public`, so putting these in `ranks` makes the indexer's intermediate state **structurally invisible** to the GraphQL API — the design's core requirement. It's a rebuildable cache (everything is derivable from `knowledge.edits`), and it lives in the same DB so the indexer can join public `members`/`editors`/`spaces`/`entities` cross-schema during aggregation.

## Verification

- `drizzle-kit migrate` applies the full chain cleanly (scaffolded with `generate --custom` so the journal + snapshot stay consistent).
- Confirmed the four tables land in `ranks` and are **absent from `public`** (0 `ranking*` tables there).

## Notes for reviewers

- **First non-`public` schema** created via Drizzle in this repo — flagging the new pattern. Alternative considered: have the ranking-indexer own its own DDL/migrations rather than the API migration chain. Went with `api/drizzle` for V1 (single migration runner, repo precedent of indexer tables living here), but happy to move it if the team prefers the indexer own its private schema.
- A few **column shapes are inferred** from the prose-level design (the dedup "update markers" on `rankings`, date storage, restriction representation) and noted in a comment. Low-risk to refine since the schema is a rebuildable cache.
- No `public` schema change, so the `0061` snapshot is an unchanged copy of `0060`'s.

### DIFF
--- a/api/drizzle/0061_create_ranks_schema.sql
+++ b/api/drizzle/0061_create_ranks_schema.sql
@@ -1,76 +1,48 @@
--- ranking-indexer working schema (GEO ranking aggregation).
---
--- These tables are the ranking-indexer's PRIVATE working cache. They live in a
--- dedicated `ranks` schema (NOT `public`) on purpose: PostGraphile introspects
--- only `public`, so nothing here reaches the GraphQL API. Every value is
--- derivable from the `knowledge.edits` stream and can be rebuilt by replay —
--- this is a cache, not a source of truth. The indexer joins the public
--- `members`/`editors`/`spaces`/`entities` tables cross-schema during
--- aggregation, which is why these sit in the same database.
---
--- NOTE: some column shapes are inferred from the design doc (prose-level) and
--- may be refined once the decode path is built — e.g. the dedup "update
--- markers" on `rankings`, date storage, and the restriction representation.
--- Because the schema is a rebuildable cache, those are low-risk to adjust.
-
-CREATE SCHEMA IF NOT EXISTS ranks;
+CREATE SCHEMA "ranks";
 --> statement-breakpoint
-
--- One row per Ranking Block entity.
-CREATE TABLE IF NOT EXISTS ranks.ranking_blocks (
-	"id" uuid NOT NULL,                            -- Ranking Block entity id
-	"space_id" uuid NOT NULL,                      -- space the block lives in (an entity can be perspectived across spaces)
+CREATE TABLE "ranks"."ranking_blocks" (
+	"id" uuid NOT NULL,
+	"space_id" uuid NOT NULL,
 	"name" text,
-	"filter" text,                                 -- query-data-block filter string (optional)
-	"start_date" timestamp with time zone,         -- optional submission window (inclusive)
+	"filter" text,
+	"start_date" timestamp with time zone,
 	"end_date" timestamp with time zone,
-	"restriction_id" uuid,                         -- aggregation restriction value entity; NULL => default "Members and editors"
+	"restriction_id" uuid,
 	"updated_at" timestamp with time zone DEFAULT now() NOT NULL,
-	-- Composite key: the same block id can be perspectived under multiple spaces.
-	CONSTRAINT "ranking_blocks_pkey" PRIMARY KEY ("id", "space_id")
+	CONSTRAINT "ranking_blocks_id_space_id_pk" PRIMARY KEY("id","space_id")
 );
 --> statement-breakpoint
-
--- One row per Rank submission entity.
-CREATE TABLE IF NOT EXISTS ranks.rankings (
-	"id" uuid NOT NULL,                            -- Rank entity id
-	"block_id" uuid,                               -- nullable until the RANK_BLOCK link arrives (partial-state model)
-	"space_id" uuid NOT NULL,                      -- submitter's personal space (a rank can be perspectived across spaces)
-	"author_address" text,                         -- submitter address resolved from the personal space
-	"rank_type" text,                              -- 'ORDINAL' | 'WEIGHTED'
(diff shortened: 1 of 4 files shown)


0xJagger:
Nice! One thing before we merge: instead of hand-writing the migration to create the schema and tables, could you define them through Drizzle's `pgSchema` and let drizzle-kit generate the migration?

We define the public schema in `gaia/api/src/services/storage/schema.ts`, maybe we could add a new `schema.ranks.ts` and define the new schema there. Something like:

```ts
// gaia/api/src/services/storage/schema.ranks.ts

import { serial, text, pgSchema } from "drizzle-orm/pg-core";
export const ranksSchema = pgSchema("ranks");

export const rankingBlocks = ranksSchema.table("ranking_blocks", {
  id: serial("id").primaryKey(),
  // ...
});

export const rankings = ranksSchema.table("rankings", {
  id: serial("id").primaryKey(),
  // ...
});

export const rankingItems = ranksSchema.table("ranking_items", {
  id: serial("id").primaryKey(),
  // ...
});

export const rankingScores = ranksSchema.table("ranking_scores", {
  id: serial("id").primaryKey(),
  // ...
});
```

Then `drizzle-kit generate` produces the `CREATE SCHEMA "ranks"` + all the table DDL for you. Keeps the schema as the source of truth so the TS definitions and the DB stay in sync (and future migrations diff cleanly against it) — manual migrations tend to drift from the schema files over time.

Docs for reference: https://orm.drizzle.team/docs/schemas

api/drizzle/0061_create_ranks_schema.sql
```diff
@@ -0,0 +1,72 @@
...
+-- markers" on `rankings`, date storage, and the restriction representation.
+-- Because the schema is a rebuildable cache, those are low-risk to adjust.
+
+CREATE SCHEMA IF NOT EXISTS ranks;
+--> statement-breakpoint
+
+-- One row per Ranking Block entity.
+CREATE TABLE IF NOT EXISTS ranks.ranking_blocks (
+	"id" uuid PRIMARY KEY NOT NULL,                -- Ranking Block entity id
+	"space_id" uuid NOT NULL,                      -- space the block lives in
```
0xJagger: We should use a composite primary key on `(id, space_id)` since the same rank can exist in multiple spaces — e.g. someone wants to create a new perspective on the same rank

api/drizzle/0061_create_ranks_schema.sql
```diff
@@ -0,0 +1,72 @@
...
+	"end_date" timestamp with time zone,
+	"restriction_id" uuid,                         -- aggregation restriction value entity; NULL => default "Members and editors"
+	"updated_at" timestamp with time zone DEFAULT now() NOT NULL
+);
+--> statement-breakpoint
+
+-- One row per Rank submission entity.
+CREATE TABLE IF NOT EXISTS ranks.rankings (
+	"id" uuid PRIMARY KEY NOT NULL,                -- Rank entity id
+	"block_id" uuid,                               -- nullable until the RANK_BLOCK link arrives (partial-state model)
```
0xJagger: Should we support submitting the same rank across different ranking blocks?

0xJagger (commented):
Looks good — left some comments. afaik these won't be a problem for v1, since ranks will most likely be created through the UI, but we may want to solve some of them now

ohohoreilly:
Addressed both review points:
- *Composite PKs:* `ranking_blocks` and `rankings` now keyed on `(id, space_id)` for cross-space perspectives (items/scores on their full tuples); indexer upserts updated to match.
- *Drizzle-owned migration:* defined the `ranks` schema + tables in `schema.ts` via `pgSchema` and regenerated the migration with drizzle-kit (replacing the hand-written one). It doesn't reach PostGraphile since that introspects the live `public` schema, not `schema.ts`.

Verified `drizzle-kit migrate` applies cleanly and the indexer e2e passes against the generated schema. Replied inline on the multi-block question (recommend v2).

api/drizzle/0061_create_ranks_schema.sql
ohohoreilly: Done — both `ranking_blocks` and `rankings` are now keyed on `(id, space_id)` (items/scores on their full tuples), so the same block/rank id can be perspectived across spaces without colliding. Updated the indexer upserts' ON CONFLICT targets to match.

api/drizzle/0061_create_ranks_schema.sql
ohohoreilly: For v1 a rank links to a single block (the RANK_BLOCK relation -> rankings.block_id). Supporting the same rank across multiple blocks would need a different model — membership would come from the rank's RANK_BLOCK relations (could be several) and the indexer would aggregate into each. I'd defer to v2 unless product wants it now. Note aggregation/dedup is already keyed per (block, space), so a join table would be the main addition.

0xJagger (approved):
lgtm! :shipit:

ohohoreilly:

ohohoreilly:

PR #726, opened 2026-06-09
feat(ranking-indexer): scaffold the rank-aggregation consumer

ohohoreilly:
## What

New `ranking-indexer` Rust crate — a Kafka consumer that turns rank submissions into the private `ranks` working schema. This is the **dependency-light foundation** of the ranking-indexer, and it's independent of the design's open questions (eligibility/scoring/projection), so it's safe to build now.

> **Stacked on #724** (`feat/rank-system-ids`) — base this PR there so the diff is just the new crate. Retarget to `main` once #724 lands. Also assumes the `ranks` schema from #725.

## What's built (compiles + tested)

- **consumer** — manual-commit `StreamConsumer` on `knowledge.edits`; decodes `HermesEdit` (prost) then the GRC2/GRC2Z payload via the `grc-20` crate (same decoder as kg-indexer)
- **storage** — idempotent upserts into `ranks.{ranking_blocks,rankings,ranking_items}` (runtime `sqlx`, builds without a DB)
- **dedup** — most-recently-updated submission per (block, personal space) — **unit-tested (4 cases)**
- **main** — per-edit loop (decode → detect → upsert); commits offsets; skips malformed edits without stalling the partition (design §10)

## Deliberately stubbed seams (focused follow-ups)

- `detect()` — the `grc_20::Op` pattern-matching for the four rank op shapes. Needs the grc-20 op API + `TYPES`-relation resolution, mirroring kg-indexer's `extract_*`. Returns empty for now (pipeline is a correct no-op until wired).
- **per-block recompute** (dedup → eligibility → scoring) and **`RANK_POSITION` projection** — gated on the design's open questions (personal-space eligibility, scoring normalization, the public-projection / atlas coordination raised in the design review).

## Verification

- `cargo build -p ranking-indexer` → clean.
- `cargo test -p ranking-indexer` → 4 dedup tests pass.

(Local build of the rdkafka layer needs `clang/libclang-dev`, `cmake`, `libcurl4-openssl-dev` — same native deps the existing rdkafka indexers require; CI already has them.)

### DIFF
--- a/ranking-indexer/src/detect.rs
+++ b/ranking-indexer/src/detect.rs
@@ -2,19 +2,77 @@
 //!
 //! The indexer keeps only the four op patterns from the design (§5) and
 //! discards everything else:
+//!   - `CreateEntity` typed `Rank`           -> a [`Ranking`]
 //!   - `CreateEntity` typed `Ranking Block`  -> a [`RankingBlock`]
-//!   - `CreateEntity` typed `Rank`           -> a [`Ranking`] (+ its items)
 //!   - `RANK_VOTES` relations                -> [`RankingItem`] rows
 //!   - `CreateRelation` of type `RANK_BLOCK`  -> a rank -> block link
 //!
-//! Type membership is read from the `TYPES` relation, matching how the rest of
-//! gaia resolves entity types. The relevant system IDs live in
-//! [`sdk::core::ids`] (`RANK_TYPE_ID`, `RANKING_BLOCK_TYPE_ID`,
-//! `RANK_VOTES_RELATION_TYPE_ID`, `RANK_BLOCK_RELATION_TYPE_ID`, …).
+//! Type membership is read from the `TYPES` relation (a `CreateRelation` of
+//! type `TYPE_RELATION_TYPE_ID` whose `to` is the type entity), matching how
+//! the rest of gaia resolves entity types.
 
+use std::collections::{HashMap, HashSet};
+use std::sync::LazyLock;
+
+use grc_20::{Id as Grc20Id, Op as Grc20Op, PropertyValue, Value as Grc20Value};
 use uuid::Uuid;
 
 use crate::models::{Ranking, RankingBlock, RankingItem};
+
+/// System IDs we match against, parsed once from their string constants.
+struct DetectIds {
+    type_relation: Uuid,
+    name_property: Uuid,
+    rank_type: Uuid,
+    ranking_block_type: Uuid,
+    rank_type_property: Uuid,
+    rank_votes_relation: Uuid,
+    rank_block_relation: Uuid,
+    vote_weighted_value_property: Uuid,
+}
+
+static IDS: LazyLock<DetectIds> = LazyLock::new(|| {
+    use sdk::core::ids::*;
+    let uid = |s: &str| Uuid::parse_str(s).expect("invalid system ID constant");
+    DetectIds {
+        type_relation: uid(TYPE_RELATION_TYPE_ID),
+        name_property: uid(NAME_PROPERTY_ID),
+        rank_type: uid(RANK_TYPE_ID),
+        ranking_block_type: uid(RANKING_BLOCK_TYPE_ID),
+        rank_type_property: uid(RANK_TYPE_PROPERTY_ID),
+        rank_votes_relation: uid(RANK_VOTES_RELATION_TYPE_ID),
+        rank_block_relation: uid(RANK_BLOCK_RELATION_TYPE_ID),
+        vote_weighted_value_property: uid(VOTE_WEIGHTED_VALUE_PROPERTY_ID),
+    }
+});
+
+fn id_to_uuid(id: &Grc20Id) -> Uuid {
+    Uuid::from_bytes(*id)
+}
+
+fn text_value(values: &[PropertyValue], property: Uuid) -> Option<String> {
+    values.iter().find_map(|pv| {
+        if id_to_uuid(&pv.property) != property {
+            return None;
+        }
+        match &pv.value {
+            Grc20Value::Text { value, .. } => Some(value.to_string()),
+            _ => None,
+        }
+    })
+}
+
+fn float_value(values: &[PropertyValue], property: Uuid) -> Option<f64> {
+    values.iter().find_map(|pv| {
+        if id_to_uuid(&pv.property) != property {
+            return None;
+        }
+        match &pv.value {
+            Grc20Value::Float { value, .. } => Some(*value),
+            _ => None,
+        }
+    })
+}
 
 /// The rank-relevant ops extracted from a single edit.
 #[derive(Debug, Default)]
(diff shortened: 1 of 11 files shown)


ranking-indexer/src/detect.rs
```diff
@@ -0,0 +1,195 @@
...
+            Grc20Op::CreateRelation(relation) => {
+                let type_id = id_to_uuid(&relation.relation_type);
+                if type_id == ids.type_relation {
+                    types_of
+                        .entry(id_to_uuid(&relation.from))
+                        .or_default()
+                        .insert(id_to_uuid(&relation.to));
+                }
+            }
+            _ => {}
```
0xJagger, line 124: Should we include the [DeleteEntity](https://docs.rs/grc-20/latest/grc_20/model/op/struct.DeleteEntity.html) and [DeleteRelation](https://docs.rs/grc-20/latest/grc_20/model/op/struct.DeleteRelation.html) as these will be emitted by the update rank sdk function? [Source](https://github.com/geobrowser/geo-sdk/blob/768409952698b680cc7035c36b5008a89b892ff8/src/ranks/update-rank.ts#L57)

ranking-indexer/src/detect.rs
```diff
@@ -0,0 +1,195 @@
...
+                        block_id: None, // set when the RANK_BLOCK link is seen (here or later)
+                        space_id,
+                        author_address: None, // resolved from the personal space during aggregation
+                        rank_type: text_value(&entity.values, ids.rank_type_property),
+                        submitted_at: None, // TODO: derive from edit/block timestamp
+                        updated_at_block: block_number,
+                        update_index: op_index as i64,
+                    });
+                } else if is_block {
+                    // NOTE: only id/space_id/name are confidently extractable today.
```
0xJagger, line 151: We don't have a function planned in the SDK to create ranking blocks. Maybe we should ask Preston whether this is something we can implement, I think having a canonical way to create ranking blocks would help both the UI and the backend

ranking-indexer/src/storage.rs
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    /// Upsert a Ranking Block.
+    pub async fn upsert_ranking_block(&self, b: &RankingBlock) -> Result<(), IndexerError> {
+        sqlx::query(
+            r#"
+            INSERT INTO ranks.ranking_blocks
+                (id, space_id, name, filter, start_date, end_date, restriction_id, updated_at)
+            VALUES ($1, $2, $3, $4, $5, $6, $7, now())
+            ON CONFLICT (id) DO UPDATE SET
```
0xJagger: Should we update this to check (id, space_id)?

0xJagger (commented):
Looking good, left some comments 🤝

ranking-indexer/src/detect.rs
0xJagger: Discard this one — I just saw that the following pr implements this

ranking-indexer/src/detect.rs
0xJagger: Edit: the `storage::replace_ranking_items` handle this by deleting all existing items for that rank right?

ranking-indexer/src/storage.rs
ohohoreilly: Done in c0b7c41 — both `upsert_ranking_block` and `upsert_ranking` now use `ON CONFLICT (id, space_id)` to match the composite PK from #725, and I dropped the now-redundant `space_id = EXCLUDED.space_id` from the SET (it's part of the key). Rebased #727/#728 on top so the whole stack is consistent.

ranking-indexer/src/detect.rs
ohohoreilly: Agreed — a canonical `createRankingBlock` in the SDK would let both the UI and this indexer rely on one op shape instead of guessing. That's exactly why the block branch here only fills id/space_id/name and leaves the window/filter/restriction fields as TODO (I didn't want to hard-code an emission shape that the SDK might land differently). I'll raise it with Preston and link the geo-sdk issue here; once it exists I'll wire the remaining fields in the eligibility branch (#727), which is where block config extraction lives.

ranking-indexer/src/detect.rs
ohohoreilly: Good catch to check. For the aggregation cache we don't actually need to process those deletes, because of how a re-submission flows through: `updateRank` re-emits the **complete** new vote set (`buildVoteOps` over all new votes), and the persist path replaces a rank's items wholesale — `replace_ranking_items` is a `DELETE ... WHERE ranking_id = $1` followed by inserts (see main.rs, "Re-submission rebuilds a rank's items, so replace per ranking"). So adds, removals, and reorders are all captured by the new `RANK_VOTES` creates alone.

The `DeleteRelation`/`DeleteEntity` ops `updateRank` emits are there to clean up the *prior* reified vote entities/relations in the canonical graph (kg-indexer's concern). They aren't actionable on our side anyway: a `DeleteRelation` op carries only the relation id (no from/to/type), and we key `ranking_items` on `(ranking_id, entity_id, space_id)` rather than relation or vote-entity id — so there's nothing to match them against.

The one case wholesale-replace can't see is a rank cleared to *zero* items or a whole-rank deletion. There's no `deleteRank` in the SDK yet (only create/update). If/when it lands it'll emit `DeleteEntity(rankId)`, and at that point I'd surface `DeleteEntity` from detect() so a deleted rank id drops the ranking + its items before recompute. Happy to add that hook defensively now if you'd prefer — flagging it as a v2 item otherwise.

0xJagger: Sounds good, thanks

0xJagger (approved):
🚀

ranking-indexer/src/detect.rs
ohohoreilly: Followed up with Preston — he pointed me at the canonical *Ranking block* type in the root space:
https://www.geobrowser.io/space/a19c345ab9866679b001d7d2138d88a1/150db6defe2344f0805afa57502e2c32

Decoding it (entity `150db6de…`, a `Type`) confirms the full 8-property ontology, and the block-config extraction in #727 already matches it 1:1 — including the parts I wasn't sure about: *Aggregation restriction* is a **relation** to an option entity (e.g. "Editors and members"), *Start/End date* are `Date`, *Filter* is `Text`, and the `Rank position` / `Aggregated rankings` output relations match what the publish stage writes. So the IDs we pinned check out against the live type.

Two things still open (tracking separately, not blockers for this PR):
- Whether a canonical `createRankingBlock` SDK builder is coming, or the UI emits raw ops to this ontology — either's fine for the indexer now that the IDs are stable.
- One ontology property, *Data source type* (relation → "Data source"), isn't read by v1 aggregation; pinned the ID for completeness in #727, confirming with Preston whether aggregation should ever branch on it.

ranking-indexer/src/detect.rs
```diff
@@ -0,0 +1,195 @@
...
+                    // perspective, and the weighted value lives on the reified vote
+                    // entity (ordinal ranks carry only the fractional index).
+                    let reified = id_to_uuid(&relation.entity_id());
+                    let weight = entity_values
+                        .get(&reified)
+                        .and_then(|vals| float_value(vals, ids.vote_weighted_value_property));
+                    out.items.push(RankingItem {
+                        ranking_id: from,
+                        entity_id: id_to_uuid(&relation.to),
+                        space_id: relation.to_space.map(|id| id_to_uuid(&id)).unwrap_or(space_id),
```
0xNeoDev: wondering if a `none` on `relation.to_space` should be treated as an error. not sure if it is expected to be different than the ranking's `space_id` cc @0xJagger

ranking-indexer/src/storage.rs
```diff
@@ -0,0 +1,136 @@
...
+        Ok(())
+    }
+
+    /// Set a rank's block link (the `RANK_BLOCK` relation may land in a later edit).
+    pub async fn set_ranking_block(
+        &self,
+        ranking_id: Uuid,
+        block_id: Uuid,
+    ) -> Result<(), IndexerError> {
+        sqlx::query("UPDATE ranks.rankings SET block_id = $2, updated_at = now() WHERE id = $1")
```
0xNeoDev: should we add `space_id` to the where clause? 

note`space_id` is in the PK https://github.com/geobrowser/gaia/pull/725/changes#diff-5b284fb2bc22e763c5a14b88fc01cd27760b1de9ada7846cae3bc1dfa88e0cdeR12

0xNeoDev (changes requested):

0xNeoDev:
left a couple comments. I'm not familiar with rankings. hope they're hopeful 😅

ranking-indexer/src/storage.rs
ohohoreilly: Yep — fixed in 4106417. `set_ranking_block` now scopes the update: `WHERE id = $1 AND space_id = $3`, with the space being the edit's space (the RANK_BLOCK relation is authored in the ranking's space). Good catch — this was the same composite-key gap as the upserts. (It threaded up the stack: the per-edit pipeline in #728 now passes `space_id` through `apply_detected_edit` to reach it.)

ranking-indexer/src/detect.rs
ohohoreilly: Good question, and you're right that it shouldn't silently default. `to_space` is the *perspective* of the ranked entity — which space that entity is being ranked in — and it's normally **not** the ranking's own (personal) space; it's wherever the target entity lives. The SDK sets it on every vote (`buildVoteOps` keys uniqueness on `(entityId, spaceId)`), so a well-formed edit always has it, and both aggregation and the published projection key on it.

So a missing `to_space` is a malformed vote we can't place in a perspective. The old `unwrap_or(space_id)` would mis-bucket it into the ranking's space and produce an aggregate row readers can't resolve. Changed it (4106417) to **skip that item** (with a `warn!`) rather than default — per-item, so one bad vote doesn't drop the rest of the edit. Didn't go as far as erroring the whole edit since that'd stall the partition over a single item. Added a unit test (`rank_vote_without_to_space_is_skipped`) covering it. cc @0xJagger in case you'd prefer hard-erroring instead.

0xNeoDev (approved):

ohohoreilly:

ohohoreilly:

PR #727, opened 2026-06-09
feat(ranking-indexer): eligibility + recompute orchestration

ohohoreilly:
## What

The ranking-indexer's **eligibility stage** plus the **per-block recompute orchestration** that drives it. Design §7, per the team's decisions in the design thread.

> **Stacked on #726** (the consumer scaffold). Base this PR there; retarget once #726 lands. Draft while the feature is in progress.

## Eligibility (`eligibility.rs`)
- **Restriction:** DAO-space block → submitter ∈ `members ∪ editors` of the block's space; personal-space block → **"All of Geo"** (admit all) for now, per Preston (narrows to verified users once verification ships).
- **Window:** submission timestamp within the block's optional `[start, end]` (absent bound = no limit).
- **Current-membership semantics:** revoked membership drops past votes on the next recompute — falls out naturally since eligibility resolves against current membership each run.
- Pure, unit-tested: `restriction_admits` / `window_admits` / `filter_eligible` (4 cases).

## Recompute orchestration (`recompute.rs`, wired into `main`)
- `affected_blocks()` — blocks whose settings changed, newly-linked blocks, and the current block of any touched rank/item (resolved from the working tables).
- `recompute_block()` — full per-block pipeline: **dedup → eligibility → [scoring stub] → [publish stub]**.
- Scoring and publish are clearly-marked stubs, gated on the open design questions (ballot normalization method; aggregate-weight property + atlas inclusion). When they land, steps 3–4 run in one transaction and the Kafka offset commit follows it.

## Storage / models
- New reads: `space_kind` (from `public.spaces.type`), `member_and_editor_spaces` (`public.members ∪ editors`), `block_id_for_ranking`, `get_ranking_block`, `get_rankings_for_block`.
- `FromRow` derived on the row models. Unknown space type → treated conservatively as DAO.

## Verification
`cargo build -p ranking-indexer` clean; `cargo test` → 8 pass (4 dedup + 4 eligibility).

## Notes
- The window check depends on `detect()` populating `submitted_at` (a TODO there); until then unknown timestamps are admitted rather than dropped.
- Full per-block recompute on every relevant edit (V1, always-correct); incremental recompute is a later optimization.

PR #728, opened 2026-06-09
feat(ranking-indexer): publish stage (RANK_POSITION projection)

ohohoreilly:
## What

The ranking-indexer's **publish stage** (design §8) — projects a block's scored aggregate into the public graph. This **completes the V1 pipeline**: `consume → decode → detect → upsert → recompute(dedup → eligibility → scoring → publish)`.

> **Stacked on #727.** Draft while the stack is in review. The stack was rebased onto the latest #724, so this branch carries the full ID set (incl. `RANK_POSITION_VALUE_PROPERTY_ID`).

## What it writes (per block, replaced atomically each recompute)
- **`RANK_POSITION` relations** — one per ranked target: `from` = block, `to` = ranked entity, `to_space_id` = the ranked perspective, ordered by a lexicographic `position` key.
- **Integer rank-position value** on each relation's reified entity (`RANK_POSITION_VALUE_PROPERTY`), block-scoped **0–100** for v1 (`round(100 * score / max_score)`).
- **`Aggregated rankings` provenance relations** — block → each contributing submission.

## Design points
- **Deterministic IDs** from `(block, entity, space)` (UUIDv5) so re-aggregation upserts in place; dropped targets are removed (the whole projection is delete-then-insert in one transaction).
- `values.id` is a text column, so the value-row UUID is serialized at the bind boundary (matches the SCORE mirror).
- Reified entities aren't inserted into `entities` (FKs are NOT VALID; readers join `block → RANK_POSITION → to_entity` and read the value by reified id) — can revisit if needed.

## Verification
`cargo build -p ranking-indexer` clean; `cargo test` → 23 pass (3 new `publish` tests: 0–100 scaling, position ordering, stable/distinct IDs).

## Notes / v2
- Integer is block-scoped (v1). Cross-ranking comparability would mean switching the aggregate measure from **sum** to a count-normalized mean — a product call, deferred.
- Position is a fixed-width lexicographic key (fine under full-replace); could swap to true fractional indexing if needed.
- Per the team: no atlas changes; cross-consumer consistency is V1 self-healing.

### DIFF
--- a/ranking-indexer/src/storage.rs
+++ b/ranking-indexer/src/storage.rs
@@ -14,6 +14,7 @@ use uuid::Uuid;
 use crate::eligibility::SpaceKind;
 use crate::error::IndexerError;
 use crate::models::{Ranking, RankingBlock, RankingItem};
+use crate::publish::{provenance_ids, RankPositionRow};
 use crate::scoring::ScoreRow;
 
 #[derive(Clone)]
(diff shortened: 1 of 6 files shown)


ohohoreilly:
Validated the full pipeline end-to-end against a real Postgres (public + `ranks` schemas). Added an opt-in integration test (`RANKING_INDEXER_E2E_DATABASE_URL`; skips in CI otherwise) plus a small refactor extracting the post-detect pipeline into `recompute::apply_detected_edit` (shared by main + the test).

Scenario: a DAO-space ranking block, two **member** submissions ranking A>B and one **non-member** submission ranking B>A. Result:

```
RANK_POSITION (block -> entity, ordered):
  A  position 0000000001  value 100  to_space=<block space>
  B  position 0000000002  value 50
Aggregated rankings provenance: 2   (non-member excluded by eligibility)
ranks.ranking_scores: A=2.0 (#1), B=1.0 (#2)
```

Confirms eligibility filtering, scoring (sum of [0.5,1] normalized), the 0–100 integer projection, perspective on `to_space_id`, deterministic IDs, and the atomic public projection all work against real SQL.

ranking-indexer/src/storage.rs
```diff
@@ -258,4 +259,98 @@ impl Storage {
...
+        let rank_position = pid(RANK_POSITION_RELATION_TYPE_ID);
+        let aggregated = pid(AGGREGATED_RANKINGS_RELATION_TYPE_ID);
+        let value_prop = pid(RANK_POSITION_VALUE_PROPERTY_ID);
+
+        let mut tx = self.pool.begin().await?;
+
+        // 1. Drop prior value rows on this block's reified RANK_POSITION entities
+        //    (before the relations they hang off are deleted).
+        sqlx::query(
+            "DELETE FROM values WHERE property_id = $1 AND entity_id IN \
```
0xJagger: Should we include space_id in these delete queries?

0xJagger (commented):
Looking good, left one small comment

ranking-indexer/src/storage.rs
ohohoreilly: Good call — done in 5df8494. Both projection-replace deletes (the `RANK_POSITION`/`Aggregated rankings` relations and the reified value rows) now carry `AND space_id = block_space_id`, and I scoped the values subquery's relation lookup too.

In practice we only ever insert the projection with `space_id = block.space_id`, so today the deletes already matched the right rows — but under the composite `(id, space_id)` key a block id can be perspectived into another space, and an unscoped delete *would* clobber that other space's projection. Scoping makes "replace this block's projection in *this* space" explicit.

Verified e2e against real Postgres: planted a same-`block_id` `RANK_POSITION` row in a second space, ran the scoped delete, and confirmed it clears the block's own-space projection (3 → 0) while the other space's row survives (1 → 1).

Related note: `recompute_block`/`get_ranking_block` still key on `block_id` alone (one row per block today, since a block is authored in one home space). If we ever do recompute the same block id per-space, those would need a space param too — flagging it but it's not exercised in v1.

0xNeoDev: yeah shouldnt be done in practice. I was thinking if adversary could bypass the geo-sdk and manually use the same rank id in their own space to overwrite this

ohohoreilly:

ohohoreilly:

Issue #738, opened 2026-06-11
ranking-indexer: block never registered when its TYPES relation + config arrive in separate edits → rankings never scored


ohohoreilly:
## Symptom
A user creates a ranking block and submits rankings, but **no RANK_POSITION system properties ever appear** on the block. The rankings are accepted but never scored/published. Does not self-heal.

## Evidence (production, 2026-06-11)
Block `b737c00c-4e64-47ee-b979-a5a900124ca2`, space `f3dab79c-b5a3-d9d1-7596-56dd5361d1c6`:

| check | result |
|---|---|
| block entity in `public.entities` | ✅ 1 |
| block has `TYPES → Ranking Block` relation in `public.relations` (type `8f151ba4…`, to `150db6de…`) | ✅ 1 |
| **registered in `ranks.ranking_blocks`** | ❌ **0** |
| rankings in `ranks.rankings` | ✅ 2 |
| ranking items | ✅ 7 |
| **scores in `ranks.ranking_scores`** | ❌ **0** |

So the submissions were ingested within seconds and recorded, but the block was never registered and scoring never ran.

## Root cause
Two interacting facts:

1. **`detect()` resolves an entity's types only from the *current edit's* ops** (`types_of` is built from `CreateRelation` ops in the edit being processed) — it never consults the already-indexed graph. So a block is only recognized when its config (Name/Filter/…) **and** its `TYPES → Ranking Block` relation appear in the **same edit**. `ranking-indexer/src/detect.rs` (~L154–209).
2. **`recompute_block()` hard-gates scoring/publish on the block being registered:**
   ```rust
   let Some(block) = storage.get_ranking_block(block_id).await? else {
       return Ok(()); // block not indexed → no scoring, no publish
   };
   ```
   `ranking-indexer/src/recompute.rs`.

When a client (e.g. the UI) creates the block entity and adds its type/config across **separate** edits, neither edit satisfies detect's single-edit condition, so the block is never registered — and every rank that links to it is silently never scored.

## Impact
Affects any block whose `TYPES → Ranking Block` relation and config aren't emitted in one atomic edit — likely common via the real UI. Rankings are accepted but produce nothing.

## Proposed fix (one of)
- In `detect()`, fall back to the **indexed graph** for an entity's types (treat an entity already typed `Ranking Block` in the KG as a block when its config is edited), rather than relying only on the current edit; **or**
- Register the block when the `TYPES → Ranking Block` relation is observed, reading its config from the KG; **or**
[...]

PR #730, opened 2026-06-10
feat(ranking-indexer): integration branch

ohohoreilly:
Integration branch consolidating the full ranking-indexer feature for a single merge into `dev`, per Jagger's suggestion — instead of landing the stacked chain PR-by-PR.

## What's folded in (6 PRs)
The linear stack and the schema branch, merged into one branch off `dev`:

- #724 — `feat(sdk)`: rank + ranking-block system IDs
- #725 — `feat(api)`: private `ranks` schema
- #726 — `feat(ranking-indexer)`: scaffold the rank-aggregation consumer
- #727 — `feat(ranking-indexer)`: eligibility + recompute orchestration
- #728 — `feat(ranking-indexer)`: publish stage (RANK_POSITION projection)
- #729 — `feat(ranking-indexer)`: containerize + k8s deploy + CI workflows

Because #724→#726→#727→#728→#729 are a linear stack, this was two merges: the `feat/ranking-indexer-deploy` tip (covers the whole chain) plus `feat/ranks-schema` (#725). Both merged cleanly, no conflicts.

## Scope (vs `dev`)
30 files: `ranking-indexer/` (crate + Dockerfile + k8s + 3 CI workflows), the `api/` drizzle `ranks` schema migration, `sdk/ids.rs`, and `Cargo.{toml,lock}`. No unrelated changes — the standalone scoring PRs (#715–#719) are intentionally **not** included; they merge to `main` on their own.

## Note
The feature branches were originally based on `main` (1 commit ahead of `dev`), so merging here also brings `dev` up to that commit. Once this lands, the six stacked PRs above can be closed.

PR #739, opened 2026-06-11
fix(ranking-indexer): recover unregistered blocks from the KG (#738)

ohohoreilly:
Draft fix for #738. Targets `main` because the ranking-indexer only exists on `main` right now (#730 was merged there, not `dev`) — **please retarget if you'd rather land it via `dev`.**

## Problem
`detect()` resolves an entity's types from the **current edit alone**, so a Ranking Block is only registered when its `TYPES → Ranking Block` relation and its config (Name/Filter/dates/restriction) arrive in the **same edit**. Real clients emit them across separate edits, so blocks were never registered — and `recompute_block()` short-circuits on a missing block, so every linked rank was **silently never scored**. In prod this was total: **0 blocks registered, 0 scores published** across 3 blocks / 18 rankings.

## Fix
When a rank links to a block that isn't in `ranks.ranking_blocks`, recover the block's config from the **indexed public graph** (`Storage::get_block_config_from_kg`) and register it before scoring. `recompute_block()` now backfills instead of bailing. Self-heals as soon as any rank for the block is processed.

- `storage.rs`: `get_block_config_from_kg()` — confirms the entity is typed `Ranking Block` in `public.relations` (also yields its home space), then reads Name/Filter/Start/End from `public.values` and the aggregation-restriction relation.
- `recompute.rs`: backfill-and-register on cache miss instead of early return.
- `tests/e2e.rs`: opt-in regression test reproducing the cross-edit scenario (type+config seeded in the KG, block unregistered, a rank arrives → block recovered + scored).

## Reviewer notes / caveats
- **Trigger is a rank edit**, not a block re-save. A pure block-config edit still produces no detected block/rank, so it won't trigger recompute. If we also want block edits to trigger it, `detect()`/`affected_blocks` would need to recognize block entities via the indexed graph (follow-up).
- **Existing backlog won't auto-reprocess**: the 3 stuck blocks' rank edits are already consumed (committed offsets). They'll heal when any *new* rank touches them; otherwise we need a one-off replay / offset reset of `ranking-indexer-v1`. (Noted on #738.)
- **Dates** are read from `values.datetime_utc`; confirm that's where kg-indexer writes the Start/End date properties (they're optional and were unset on the blocks I inspected).
- The e2e test is opt-in (`RANKING_INDEXER_E2E_DATABASE_URL`), like the existing one — it won't run on generic CI. Verified `cargo fmt` + `cargo clippy --all-targets -- -D warnings` clean locally; couldn't run the e2e itself here (no `ranks`-schema DB).

### DIFF
--- a/ranking-indexer/src/storage.rs
+++ b/ranking-indexer/src/storage.rs
@@ -7,6 +7,7 @@
 
 use std::collections::HashSet;
 
+use chrono::{DateTime, Utc};
 use sqlx::postgres::PgPoolOptions;
 use sqlx::PgPool;
 use uuid::Uuid;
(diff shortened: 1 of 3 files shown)


ohohoreilly:
Scope note: this PR is intentionally minimal — it fixes the observed case (block created cross-edit → recover from KG → score) and self-heals it. The **root cause** (detect() resolving types from a single edit, not indexed state) and its three residual edge cases — rank-before-type race, post-registration config updates, and block edits not triggering recompute — are tracked separately in #740.

ranking-indexer/src/storage.rs
```diff
@@ -95,6 +96,84 @@ impl Storage {
...
+            Option<DateTime<Utc>>,
+            Option<DateTime<Utc>>,
+        );
+        let (name, filter, start_date, end_date): ConfigRow = sqlx::query_as(
+            "SELECT \
+               max(text)         FILTER (WHERE property_id = $2) AS name, \
+               max(text)         FILTER (WHERE property_id = $3) AS filter, \
+               max(datetime_utc) FILTER (WHERE property_id = $4) AS start_date, \
+               max(datetime_utc) FILTER (WHERE property_id = $5) AS end_date \
+             FROM values WHERE entity_id = $1",
```
0xJagger: We should a include `space_id` here as different users may define perspectives on the same rank

ranking-indexer/src/storage.rs
```diff
@@ -95,6 +96,84 @@ impl Storage {
...
+        .bind(pid(ids::RANK_FILTER_PROPERTY_ID))
+        .bind(pid(ids::RANK_START_DATE_PROPERTY_ID))
+        .bind(pid(ids::RANK_END_DATE_PROPERTY_ID))
+        .fetch_one(&self.pool)
+        .await?;
+
+        // Aggregation restriction is a relation (as detect() collects it).
+        let restriction: Option<(Uuid,)> = sqlx::query_as(
+            "SELECT to_entity_id FROM relations \
+             WHERE from_entity_id = $1 AND type_id = $2 LIMIT 1",
```
0xJagger: Same here

0xJagger (commented):
Looking good, left some small comments

ranking-indexer/src/storage.rs
ohohoreilly: Good catch — done in the latest push. Scoped the config-values query to the block's home `space_id` (resolved from the `TYPES → Ranking Block` relation).

ranking-indexer/src/storage.rs
ohohoreilly: Done — the aggregation-restriction query is now scoped to the block's `space_id` too.

0xJagger (approved):

ranking-indexer/src/storage.rs
```diff
@@ -95,6 +96,89 @@ impl Storage {
         Ok(block)
     }
 
+    /// Reconstruct a Ranking Block's config from the indexed public graph.
+    ///
+    /// `detect()` only registers a block when its `TYPES -> Ranking Block`
+    /// relation and its config (Name/Filter/dates/restriction) arrive in the
+    /// *same* edit, because it resolves an entity's types from the current edit
+    /// alone. Real clients emit the type and the config across separate edits,
+    /// so the block is never registered and every rank linked to it is silently
```
0xNeoDev, line 105: could they be emitted in opposite order as expected here? is that case handled

ohohoreilly: Order doesn't matter here — `get_block_config_from_kg` reads the *current* indexed graph (the type relation + config values as they stand now), not a single edit, so whichever of the type/config edits arrived first, the block is reconstructed correctly as long as both are present when a rank triggers recovery. I clarified the doc comment to say this explicitly.

The cases that *are* order-sensitive — a rank seen before the block's type is indexed, or the config being edited *after* the block is already registered (recompute then reads the stored row, not the KG) — aren't handled here by design; they're the residual root-cause gaps tracked in #740. This PR is the minimal unblock.

PR #742, opened 2026-06-12
ci: add gaia-v2 manual build/deploy workflows (GEO-664)

0xJagger:
## Why

GitHub only lists `workflow_dispatch` workflows in the Actions UI when the workflow file exists on the **default branch**. These two power the gaia-v2 (chain-migration) bring-up, which deploys from the `staging` branch (snapshot of `defi-wonderland/gaia#feat/governance-v2-contract-migration`).

## What

- `build-v2-images.yml` — manual matrix build+push of the 10 gaia-v2 images to `registry.digitalocean.com/geo`, **SHA tags only, never `:latest`** — nothing running in prod can pull these
- `deploy-v2.yml` — manual single-service deploy, strictly scoped to the `gaia-v2` namespace, with two guards: the image tag must exist in the registry, and manifests containing unfilled `REPLACE_WITH_*` placeholders refuse to deploy

## Safety

- Dispatch-only: no `push`/`pull_request` triggers — merging this changes nothing by itself, and it doesn't match any existing prod workflow's path filters
- Reuses the existing `DIGITALOCEAN_ACCESS_TOKEN` / `DIGITALOCEAN_CLUSTER_NAME` secrets
- **When dispatching, select the `staging` branch as the ref** — the run uses that branch's code and `k8s/v2` manifests. Dispatched against `main` itself: the build produces unused SHA-tagged images, the deploy fails fast on missing `v2/` manifests

Part of GEO-664 — runbook: see the GEO-664 deploy doc (Steps 11–12).

### DIFF
--- a/.github/workflows/build-v2-images.yml
+++ b/.github/workflows/build-v2-images.yml
@@ -1,0 +1,87 @@
+name: Build gaia-v2 images (manual)
+
+# Builds and pushes all images needed for the gaia-v2 (chain-migration) stack.
+# Build-and-push only — deploys stay manual per the GEO-664 runbook (Step 12:
+# sed the image tag into the k8s/v2 manifests and kubectl apply).
+#
+# Images are tagged with the full commit SHA and a short 8-char SHA; use either
+# in the manifests.
+
+on:
+  workflow_dispatch:
+
+concurrency:
+  group: build-v2-images
+  cancel-in-progress: false
+
+env:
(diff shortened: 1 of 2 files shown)


0xJagger:
Closing — we'll dispatch these via `gh workflow run build-v2-images.yml --repo geobrowser/gaia --ref staging` instead (API resolves the workflow by filename on the ref), so they don't need to live on main.

Issue #740, opened 2026-06-11
ranking-indexer: resolve entity types from indexed state, not just the current edit (root cause of #738)


ohohoreilly:
## Background
#738 / #739 fixed the observed case (a Ranking Block created across edits, then ranks arrive → recover the block from the KG and score it). But #739 is a targeted fix at the *registration* boundary; the **root cause** remains and leaves edge cases unfixed.

## Root cause
`detect()` decides what an entity is using **only the current edit's ops** — it builds `types_of` from `CreateRelation` ops in that one edit and never consults the indexed graph (`ranking-indexer/src/detect.rs`). So an entity is only recognized as a Ranking Block (or Rank) when its `TYPES` relation is co-located with its config/data in a single edit. Real clients routinely split these across edits.

## Residual gaps NOT covered by #739
1. **Ordering / race — ranks before the block's type is indexed.** #739's backfill only runs on a *rank edit* and only succeeds if the block's `TYPES → Ranking Block` relation is already in `public` at that moment (i.e. `kg-indexer` has caught up). If ranks are processed first, those ranks stay unscored until *another* rank for the block arrives later — the earlier ones aren't retried.
2. **Config updates after registration.** Once a block is in `ranks.ranking_blocks`, `recompute_block()` reads the stored row and never re-checks the KG. A later cross-edit change to filter/restriction/name is missed by `detect()` and not picked up by the backfill (the block isn't *missing*) → scoring silently uses stale config.
3. **Block edits don't trigger recompute.** Only rank edits drive `recompute_block`. A pure block-config edit produces no detected block/rank, so `affected_blocks` is empty and nothing recomputes (also why "re-save the block" isn't a workaround).

## Proposed root-cause fix
Resolve identity/type from **indexed state**, not a single edit:
- In `detect()` (or a pre-pass in the consumer), resolve an entity's types by **unioning the current edit's type relations with the entity's already-indexed types** from the graph. Then a block is recognized on *any* edit touching it, regardless of co-location.
- Add **block entities touched by an edit to `affected_blocks`** (look them up in the KG / `ranks`) so config edits re-trigger scoring.
- On recompute, **refresh the block's config from the KG** instead of trusting the stored `ranking_blocks` row, so updates propagate.

This subsumes #739's backfill and closes gaps 1–3.

## Notes
- Adds a KG read into the consume path (per-edit type resolution) — weigh the cost; can be scoped to entities that appear in rank/block-relevant ops.
[...]

PR #745, opened 2026-06-12
feat(ranking-indexer): recompute aggregates on space membership events

0xJagger:
## Problem

Ranks submitted by someone who isn't a member of the space are excluded from the ranking block's aggregate — and stay excluded even after they become a member. The ranking-indexer only consumes `knowledge.edits`, and membership is read at recompute time from the kg-indexer-maintained `members`/`editors` tables, so a membership change never triggers a recompute. Reacting to membership events while reading those tables directly would also race the kg-indexer's consumer group.

## Changes

- The indexer now also consumes `space.membership` (`ROLE_GRANTED`/`ROLE_REVOKED`, dispatched on the `event-type` header — same logic as the kg-indexer) and maintains its own view of the space registry in `ranks.members`/`ranks.editors`. Eligibility reads this view instead of the public tables, eliminating the cross-indexer race.
- Each membership event updates the view, then recomputes only the blocks in the affected space that hold a submission from that member's personal space. Becoming a member/editor integrates previously-excluded ranks; a revoke drops them.
- Migration `0062` creates the view tables and seeds them from the public `members`/`editors`, so the switch-over is a semantic no-op.
- Membership messages follow the same drop-on-error policy as edits — one bad message never stalls the partition.
- Fixes a latent bug in the cross-edit KG-recovery e2e test: `relations.id` was bound as text instead of uuid, so the test failed against any real database (never caught — the e2e suite is opt-in and CI doesn't run it).

Cross-topic ordering between `knowledge.edits` and `space.membership` is undefined, but converges: whichever event lands last triggers a recompute that reads the consistent local view.

## Notes for deploy

- Aggregates that are *already* stale won't heal on their own — a block recomputes on the next event that touches it. The few affected production blocks need a one-off recompute after deploy.
- `SPACE_LEFT` is emitted by the pipeline but intentionally not handled, for parity with the kg-indexer (which also ignores it). If voluntary leaves should affect membership, both indexers should pick that up together.

## Testing

- Unit tests for the event-type dispatch and event → change mapping.
- e2e (live Postgres): non-member's rank excluded → `ROLE_GRANTED(MEMBER)` integrates it → `ROLE_REVOKED(MEMBER)` drops it → same cycle for the editor role. All 3 e2e tests pass against a fresh DB with all migrations applied.

### DIFF
--- a/api/drizzle/0062_ranks_members_editors.sql
+++ b/api/drizzle/0062_ranks_members_editors.sql
@@ -12,6 +12,7 @@ CREATE TABLE "ranks"."members" (
 --> statement-breakpoint
 CREATE INDEX "ranks_editors_space_id_idx" ON "ranks"."editors" USING btree ("space_id");--> statement-breakpoint
 CREATE INDEX "ranks_members_space_id_idx" ON "ranks"."members" USING btree ("space_id");--> statement-breakpoint
+CREATE INDEX "rankings_space_id_idx" ON "ranks"."rankings" USING btree ("space_id");--> statement-breakpoint
 -- Seed the ranking-indexer's membership view from the kg-indexer-maintained
 -- public tables. Eligibility reads the public tables until this migration runs,
 -- so the copy makes the switch-over a semantic no-op; only events consumed
--- a/api/drizzle/0062_ranks_members_editors.sql
+++ b/api/drizzle/0062_ranks_members_editors.sql
@@ -1,0 +1,24 @@
+CREATE TABLE "ranks"."editors" (
+	"member_space_id" uuid NOT NULL,
+	"space_id" uuid NOT NULL,
+	CONSTRAINT "editors_member_space_id_space_id_pk" PRIMARY KEY("member_space_id","space_id")
+);
+--> statement-breakpoint
+CREATE TABLE "ranks"."members" (
+	"member_space_id" uuid NOT NULL,
+	"space_id" uuid NOT NULL,
+	CONSTRAINT "members_member_space_id_space_id_pk" PRIMARY KEY("member_space_id","space_id")
+);
+--> statement-breakpoint
+CREATE INDEX "ranks_editors_space_id_idx" ON "ranks"."editors" USING btree ("space_id");--> statement-breakpoint
+CREATE INDEX "ranks_members_space_id_idx" ON "ranks"."members" USING btree ("space_id");--> statement-breakpoint
+-- Seed the ranking-indexer's membership view from the kg-indexer-maintained
+-- public tables. Eligibility reads the public tables until this migration runs,
+-- so the copy makes the switch-over a semantic no-op; only events consumed
+-- after deploy diverge.
+INSERT INTO "ranks"."members" ("member_space_id", "space_id")
+SELECT "member_space_id", "space_id" FROM "public"."members"
+ON CONFLICT DO NOTHING;--> statement-breakpoint
+INSERT INTO "ranks"."editors" ("member_space_id", "space_id")
+SELECT "member_space_id", "space_id" FROM "public"."editors"
+ON CONFLICT DO NOTHING;
--- a/api/drizzle/meta/0062_snapshot.json
+++ b/api/drizzle/meta/0062_snapshot.json
@@ -1,5 +1,5 @@
 {
-  "id": "9acc6507-6328-4fbe-85bb-f62567a4b6be",
+  "id": "1e70dfc8-9e9e-4392-83b1-cb0798aa1f53",
   "prevId": "66ae6367-a3a0-4a26-9998-a7e9914df646",
   "version": "7",
   "dialect": "postgresql",
@@ -1711,6 +1711,21 @@
               "asc": true,
               "nulls": "last"
             },
+            {
+              "expression": "space_id",
+              "isExpression": false,
+              "asc": true,
+              "nulls": "last"
+            }
+          ],
+          "isUnique": false,
+          "concurrently": false,
+          "method": "btree",
+          "with": {}
+        },
+        "rankings_space_id_idx": {
+          "name": "rankings_space_id_idx",
+          "columns": [
             {
               "expression": "space_id",
               "isExpression": false,
--- a/api/drizzle/meta/0062_snapshot.json
+++ b/api/drizzle/meta/0062_snapshot.json
@@ -1,0 +1,3793 @@
+{
+  "id": "9acc6507-6328-4fbe-85bb-f62567a4b6be",
+  "prevId": "66ae6367-a3a0-4a26-9998-a7e9914df646",
+  "version": "7",
+  "dialect": "postgresql",
+  "tables": {
+    "public.app_webhooks": {
+      "name": "app_webhooks",
+      "schema": "",
+      "columns": {
+        "id": {
+          "name": "id",
+          "type": "uuid",
+          "primaryKey": true,
+          "notNull": true,
+          "default": "gen_random_uuid()"
+        },
+        "app_name": {
+          "name": "app_name",
+          "type": "text",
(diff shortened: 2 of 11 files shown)


ohohoreilly:
## Review

Nice work — this is the right design. Owning a local view of the registry instead of reading the kg-indexer's tables genuinely eliminates the cross-consumer-group race rather than papering over it, the affected-blocks query is correctly space-scoped, and the e2e coverage is exactly what I'd want: the full grant → revoke lifecycle for *both* roles, plus the view-table emptiness assertion at the end. The convergence argument in the description is the right frame and the steady-state code delivers it. Also glad the uuid-vs-text bind in the cross-edit e2e got caught and fixed here.

Reading it against the consumer config surfaced two deploy-time hazards and one policy question worth settling before merge. All three are about the transition, not the steady state.

### 1. First deploy replays the entire `space.membership` topic — and publishes while it replays

The consumer group already exists with offsets for `knowledge.edits` only. There are no committed offsets for the newly subscribed topic, and the consumer sets `auto.offset.reset = earliest` (`consumer.rs`), so the first deploy replays the full retained history of `space.membership` per partition.

The view itself converges — seed + in-order replay ends at current state, and the replay even covers the seed→deploy gap, which is a nice property. The problem is that `apply_membership_event` recomputes **and publishes** per event: replaying someone's historical revoke temporarily drops them from the view and republishes the block's `RANK_POSITION` projection without their ranks, until their later re-grant replays. So during catch-up the public aggregates flap through historical states — plus one recompute+publish per historical membership event of load.

Options, in rough preference order:
- **(a)** pre-commit the membership topic offsets to latest for `ranking-indexer-v1` before the new version goes live — the 0062 seed already captures "now", so history adds nothing;
- **(b)** fresh/secondary handling that initializes membership offsets to a timestamp just before the migration ran;
- **(c)** gate publishing until the membership partitions reach their high-water mark.

### 2. Deploy order is load-bearing, and the failure mode is silent

`member_and_editor_spaces` now reads `ranks.members`/`ranks.editors` unconditionally. If the new image starts before 0062 has been applied, every recompute fails with `relation "ranks.members" does not exist` — and the existing drop-on-error policy in `main.rs` commits the offset and **permanently skips each edit**. A reversed deploy order doesn't crash-loop loudly; it silently eats every rank edit until someone notices.

At minimum this deserves a bolded line in the deploy notes (migration strictly before rollout). Ideally: a startup check that the view tables exist before subscribing — fail fast → restart loop → offsets untouched, which turns a silent data-loss mode into a visible ordering mistake.

### 3. Drop-on-error hurts more on this topic than on edits

I know this is deliberate parity with the edits policy, and the comment in `main.rs` is honest about the consequence — but the asymmetry is worth weighing. A dropped edit loses one rank. A `ROLE_REVOKED` dropped on a transient DB blip leaves a revoked member's ranks counting **indefinitely** — nothing heals it except a later role change for the same pair. Revocation is the abuse-response path, so it's the event we least want to lose.

The fix shape is cheap here: parse errors are genuinely poison (skip is right), but `apply_membership_event` only fails on DB errors — match on the variant and don't commit on `Database` (retry with backoff or exit; the writes are idempotent upserts/deletes so redelivery converges). Happy for this to be a follow-up that fixes the edits path at the same time, since the policy is shared.

### 4. SPACE_LEFT will warn on every occurrence

The pipeline emits `SPACE_LEFT` on this topic (`hermes-pipeline/src/emit.rs:377`), so the "unknown membership event type" path logs a warn for a known, expected event — same signature as genuine corruption. One extra match arm to explicitly ignore (or `debug!`) it keeps the warn level meaningful. Fine to keep it unhandled semantically per the parity note — just don't let it look like an error.

### Minor

- `blocks_with_rankings_from` is correctly space-scoped, but `recompute_block` downstream still resolves the block by `id` alone (pre-existing — same family as the #740 edge cases), so the scoping guarantee stops at the query.
- The new query filters `ranks.rankings` by `space_id` with no supporting index. Fine at current scale; cheap to add if membership churn grows.
- 0062 is missing a trailing newline.

ohohoreilly (changes requested):
Overall: steady-state design and tests are genuinely good — 1 and 2 are the ones I'd want resolved (or explicitly handled in the deploy plan) before merge, since they only bite at rollout, which is exactly when nobody's looking at logs yet.

ohohoreilly:
Pushed the review follow-ups directly to the branch rather than leaving them as homework — `c229944`. Walking through what landed, in the order of the review comment:

### 1. Membership history replay → custom rebalance hook

`consumer.rs` now has a `RankingConsumerContext` that overrides `ConsumerContext::rebalance`: on partition assignment, any `space.membership` partition with **no committed offset** gets its start position set to `Offset::End` before the assign happens. Edits partitions are untouched, and if the committed-offset lookup fails it falls back to `auto.offset.reset` (full replay) — a replay converges and only flaps published aggregates, whereas wrongly skipping ahead would silently lose events, so that's the right failure direction.

I went with the rebalance hook instead of pre-committing offsets at startup because the deployment uses the default RollingUpdate strategy — the old pod can still hold the group when the new one boots, which makes a pre-subscribe commit unreliable. One caveat, documented on the struct: if the group's committed offsets ever *expire* (inactive past the broker's `offsets.retention.minutes`), membership resumes at latest and the view needs a reseed (re-run the 0062 seed inserts). Same operational posture as the migration itself, so I think that's acceptable — flagging it so it's a known property, not a surprise.

### 2. Deploy order → fail-fast preflight

`Storage::check_membership_view()` probes `ranks.members` / `ranks.editors` at boot, before subscribing. Deploying the image before 0062 is applied is now an explicit "has migration 0062_ranks_members_editors been applied?" crash loop with offsets untouched, instead of sporadic per-message failures. Deploy order still matters — migration first — but getting it wrong is now loud and free.

### 3. Error policy → transient retries, poison skips

Added `IndexerError::is_poison()` (only `Decode` qualifies) and a `with_transient_retry` helper in the consume loop: poison → skip + commit, exactly as before; transient (DB/Kafka) → 1s/2s/4s backoff, then **exit without committing** so Kafka redelivers and the idempotent writes converge. Applied to both topics — so this also closes the pre-existing edits-path hole where a transient DB blip permanently dropped an edit. The §10 "one bad message never stalls the partition" property is preserved: malformed input is still skipped; it's only genuine infrastructure failure that now halts instead of silently losing a revocation.

### 4. SPACE_LEFT → quiet, explicit skip

`parse_membership_event` returns `Result<Option<…>>` now — `Ok(None)` for SPACE_LEFT (debug log, still intentionally unhandled for kg-indexer parity), errors reserved for genuinely unknown types. Warn-level stays meaningful.

### Minors

- `rankings_space_id_idx` added for the `blocks_with_rankings_from` query — regenerated 0062 through drizzle-kit (schema.ts + snapshot + journal consistent; seed SQL re-appended; trailing newline fixed in the bargain).
- Left the unscoped `recompute_block` lookup alone — that's the pre-existing `(id, space_id)` issue and deserves its own PR with its own tests.

### Verification

29 unit tests pass (dispatch test extended with a SPACE_LEFT case); all 3 e2e tests pass against a fresh Postgres 18 with every migration applied via `drizzle-kit migrate`, which exercises the regenerated 0062 end to end; clippy `-D warnings` clean; api `tsc` + biome clean.

The one piece with no automated coverage is the rebalance hook itself — it needs a real broker. Staging *is* the first-deploy scenario it exists for, so the staging rollout doubles as its test: expect one `No committed offset for membership partition — starting at latest` info line per partition on first boot, and silence on every boot after. If you'd rather see a broker-level integration test for it before merge, happy to look at wiring one against the compose Kafka.

ohohoreilly (approved):
Feel free to push back on any of this — it's your PR and none of these change the core design, which I think is right.

PR #748, opened 2026-06-15
fix(proposal-executor): match any AddMember in allowlisted spaces

0xyaco:
Remove the target_id = proposed_by detection filter. proposed_by is always the space (never the individual proposer), so the clause could never match a real membership request and detection always returned zero. The bot now auto-accepts any untouched, in-window, single-AddMember Fast proposal in an allowlisted space; self-service vs editor-initiated adds are not distinguished.


0xJagger (approved):
lgtm 🤝

PR #751, opened 2026-06-16
chore: main to dev

0xNeoDev:

PR #754, opened 2026-06-16
chore: main → dev (drizzle 0062 collision resolved)

0xNeoDev:
Replaces #751 (which couldn't be resolved in place — its head was `main`). Cut from `main` with `dev` merged in, so it contains every commit from both branches; `main` is untouched.

## The conflict
`main` and `dev` both branched from `0061_create_ranks_schema` and each applied a **different** migration under index **62** — to **different databases**:

| idx 62 | branch | applied on | touches |
|---|---|---|---|
| `ranks_members_editors` | main | **prod** | `ranks` schema (members/editors) |
| `context_aware_diff_columns` | dev | **dev** | `public.relation_versions` / `value_versions` |

## Resolution
- **ranks stays `0062`**; **context_aware renumbered to `0063`** (regenerated via drizzle-kit so its snapshot stacks on the ranks snapshot — not hand-renamed).
- **Both made idempotent**, because each is already applied on one DB and must re-run there as a no-op:
  - ranks: `CREATE TABLE/INDEX IF NOT EXISTS` + seed `ON CONFLICT DO NOTHING`
  - context: `ADD COLUMN IF NOT EXISTS` (its indexes were already `IF NOT EXISTS`)
- **Both `when` bumped just above the newest pre-merge migration** (context's original `1781536970767`), kept **monotonic** (`62=…768`, `63=…769`). drizzle-migrate applies by a fixed `created_at` high-water mark, so this makes every environment re-run **both** and converge:
  - **dev** (has context): ranks **creates**, context **no-ops**.
  - **prod** (has ranks): ranks **no-ops**, context **creates**.
  - **fresh CI**: both apply in order.
- `schema.ts` auto-merged to include both feature sets.

## Verification
- `drizzle-kit generate` → **no further schema changes** (snapshot chain consistent).
- `api` `tsc --noEmit` clean.
- Branch contains all of `main` and all of `dev` (`rev-list` = 0).

## ⚠️ Merge with a merge commit, NOT squash
`dev-sync-check.yml` requires `dev` to contain every `main` commit by SHA; squashing would fail it on the next push.


0xJagger (approved):

ohohoreilly (approved):
LGTM 🚀

PR #755, opened 2026-06-16
fix(ranking-indexer): populate from_space_id and register projected entities

0xJagger:
## Summary

Two correctness fixes to the public-graph projection the ranking-indexer writes (RANK_POSITION + Aggregated rankings relations), both in `replace_rank_position_projection`.

### 1. Populate `from_space_id` on projected relations
The projected relations left `from_space_id` NULL, unlike every other relation in `public.relations` (the kg-indexer always sets it). The `from` entity is the block, which lives in the block's home space, so `from_space_id` is set to `block_space_id` — the same value already bound to `space_id`. The projection-replacement DELETE mirrors the new column.

### 2. Register projected entities so the API resolves them
The projection wrote `relations` and rank-position `values` rows keyed on **reified entity ids**, but never inserted those ids into the `entities` table. `entities` is the source of truth the API's `entity(id)` resolves against, so the reified entity carrying the rank value had no row — `entity(id)` returned `null` and the value (e.g. the `100` for the top-ranked entity) never surfaced in the graph or the geobrowser.

Mirroring the kg-indexer, we now register both the relation id and its reified entity id (for RANK_POSITION rows and provenance relations), upserting them into `entities` within the projection transaction (`ON CONFLICT` bumps `updated_at` on re-aggregation). Block provenance (`created_at` / `*_block`) is threaded from the triggering edit or membership event via a new `BlockMeta` carried on `DetectedEdit` and into `recompute_block`.

## Testing
- `cargo test -p ranking-indexer`, `cargo clippy --all-targets` pass.
- New e2e assertions: `from_space_id` carries the block's home space, and no projected reified entity is left orphaned in `entities`. (e2e tests require `RANKING_INDEXER_E2E_DATABASE_URL`.)

## Notes
- Fixes **new** projections. Reified entities written before this change get their `entities` row on the next recompute of their block (resubmission, membership change, or a backfill).
- The stricter `from_space_id` predicate on the replacement DELETE won't match pre-fix rows whose `from_space_id` is NULL; deterministic relation ids mean re-aggregation upserts cleanly via the PK rather than duplicating.

### DIFF
--- a/ranking-indexer/src/detect.rs
+++ b/ranking-indexer/src/detect.rs
@@ -18,7 +18,7 @@ use chrono::{DateTime, Utc};
 use grc_20::{Id as Grc20Id, Op as Grc20Op, PropertyValue, Value as Grc20Value};
 use uuid::Uuid;
 
-use crate::models::{Ranking, RankingBlock, RankingItem};
+use crate::models::{BlockMeta, Ranking, RankingBlock, RankingItem};
 
 /// System IDs we match against, parsed once from their string constants.
 struct DetectIds {
@@ -120,6 +120,8 @@ pub struct DetectedEdit {
     pub items: Vec<RankingItem>,
     /// `(ranking_id, block_id)` links from `RANK_BLOCK` relations.
     pub block_links: Vec<(Uuid, Uuid)>,
+    /// Block provenance of the edit, recorded on entities the recompute mints.
+    pub meta: BlockMeta,
 }
 
 impl DetectedEdit {
@@ -146,7 +148,13 @@ pub fn detect(
 ) -> DetectedEdit {
     let ids = &*IDS;
     let submitted_at = DateTime::<Utc>::from_timestamp(block_timestamp, 0);
-    let mut out = DetectedEdit::default();
+    let mut out = DetectedEdit {
+        meta: BlockMeta {
+            number: block_number,
+            timestamp: block_timestamp,
+        },
+        ..DetectedEdit::default()
+    };
 
     // Pass 1: index entity property-values and resolve TYPES membership, and
     // collect the rank relations (votes + block links).
--- a/ranking-indexer/src/membership.rs
+++ b/ranking-indexer/src/membership.rs
@@ -10,6 +10,7 @@ use hermes_schema::pb::membership::{HermesRoleGranted, HermesRoleRevoked, Member
 use uuid::Uuid;
 
 use crate::error::IndexerError;
+use crate::models::BlockMeta;
 use crate::recompute::recompute_block;
 use crate::storage::Storage;
 
@@ -35,6 +36,20 @@ pub struct MembershipChange {
     pub space_id: Uuid,
     pub member_space_id: Uuid,
     pub kind: ChangeKind,
+}
+
+/// Block provenance of the event, recorded on entities a triggered recompute
+/// mints. Defaults to zero when meta is absent, matching the edit path.
+fn block_meta(event: &MembershipEvent) -> BlockMeta {
+    let meta = match event {
+        MembershipEvent::RoleGranted(e) => e.meta.as_ref(),
+        MembershipEvent::RoleRevoked(e) => e.meta.as_ref(),
+    };
+    meta.map(|m| BlockMeta {
+        number: m.block_number as i64,
+        timestamp: m.created_at as i64,
+    })
+    .unwrap_or_default()
 }
 
 fn ids(space_id: &[u8], member_space_id: &[u8]) -> Result<(Uuid, Uuid), IndexerError> {
@@ -86,6 +101,7 @@ pub async fn apply_membership_event(
     storage: &Storage,
 ) -> Result<(), IndexerError> {
     let change = change_for(event)?;
+    let meta = block_meta(event);
 
     match change.kind {
         ChangeKind::AddMember => {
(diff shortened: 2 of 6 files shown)


ohohoreilly:
## Review

Really nice fix — the entity-registration half especially. Mirroring the kg-indexer's `UNNEST` + `ON CONFLICT (id) DO UPDATE` shape (down to reusing `$2`/`$3` so `created_at == updated_at` on first insert and re-aggregation only bumps `updated_at`) is exactly the right move, and threading `BlockMeta` from both the edit and membership paths through `recompute_block` is clean. Diagnosing that the reified entity carrying the rank value had no `entities` row — so `entity(id)` returned null and the value never surfaced — is a good catch; that's the kind of thing that's invisible until someone goes looking for why a `100` isn't showing up in the browser. `from_space_id = block_space_id` on both relation types is correct too (the `from` is the block, which lives in the block's home space).

One thing I think needs resolving before merge, though, and it bites exactly the blocks this PR exists to repair.

### The `from_space_id` predicate on the replacement DELETE collides instead of upserting

Step 2 gained `AND from_space_id = $4` (bound to `block_space_id`):

```sql
DELETE FROM relations WHERE from_entity_id = $1 AND type_id = ANY($2) AND space_id = $3 AND from_space_id = $4
```

Pre-fix projection rows have `from_space_id IS NULL`, so `NULL = block_space_id` is never true and **the DELETE skips every one of them**. The relation ids are deterministic v5 (`derive("rank_position:{block}:{entity}:{space}")`, `provenance_ids(...)`) and `relations.id` is the PK, so step 4/5 then re-`INSERT INTO relations` with those same ids — and there's no `ON CONFLICT` on the relations inserts (only the new `entities` insert has one). So it's a duplicate-key violation, the txn aborts, and `replace_rank_position_projection` returns a `Database` error.

This is where the "deterministic relation ids mean re-aggregation upserts cleanly via the PK rather than duplicating" note doesn't hold — there's no upsert on the relations rows, so instead of upserting it collides. And with #745 already in `main`, a `Database` error isn't poison: it's retried 4× and then `std::process::exit(1)` → restart → same block redelivered → same collision → crash loop. The trigger is the first edit or membership event touching any block that already has a pre-fix (NULL-`from_space_id`) projection — i.e. the exact population the Notes section says we want to re-aggregate.

**Simplest fix that also makes the transition self-heal:** drop `AND from_space_id = $4` (and its bind) from the DELETE. It's already scoped by `from_entity_id = block_id AND space_id = block_space_id`, and we always write `from_space_id` equal to `space_id` here, so the predicate is redundant for new rows and harmful for old ones. Without it the DELETE clears both NULL and populated rows, the inserts never collide, and pre-fix blocks repair on first recompute. (Could also add `ON CONFLICT (id) DO UPDATE` to the relation inserts as belt-and-braces, but removing the predicate is the actual fix.)

Quick way to size the blast radius before deploy:
```sql
SELECT count(*) FROM relations
WHERE type_id IN ('<RANK_POSITION>', '<AGGREGATED_RANKINGS>') AND from_space_id IS NULL;
```
If that's nonzero, it fires on the first re-aggregation after deploy.

### Test gap that hides it

The new assertions (`from_space_id` populated, no orphaned entities) are good and correct. But every e2e test starts with `DELETE FROM relations WHERE space_id = $1`, so they always run clean and never exercise *a pre-existing NULL-`from_space_id` row being re-aggregated* — which is the path that collides, which is why CI is green. A regression that seeds a relation with the deterministic id and `from_space_id = NULL`, then recomputes and asserts success + the row now carrying `from_space_id`, would both catch this and guard the transition going forward.

### Minor

`block_meta` defaulting to `(0,0)` when a membership event has no meta stamps `created_at = "0"` on any entity first registered that way (the `ON CONFLICT` only refreshes `updated_at`, so it sticks). Consistent with the edits path's `(0,0)` default and purely cosmetic — not blocking, just noting it.

ohohoreilly (changes requested):
Everything else looks right to me — the values DELETE still cleans old rows fine since its subquery doesn't filter `from_space_id`, and there are no intra-run id collisions in the entities `UNNEST`. Just the DELETE predicate + a regression test and I think this is good to go.

0xJagger:
> Everything else looks right to me — the values DELETE still cleans old rows fine since its subquery doesn't filter `from_space_id`, and there are no intra-run id collisions in the entities `UNNEST`. Just the DELETE predicate + a regression test and I think this is good to go.

Good catch, fixed in [f336532](https://github.com/geobrowser/gaia/pull/755/commits/f33653213ccd05c2e5bb8632a7969afc2d69559d).

You're right that the from_space_id predicate on the step-2 DELETE was worse than a missed cleanup. Pre-fix projection rows have from_space_id IS NULL, so AND from_space_id = block_space_id never matched them — the DELETE skipped them, and since the relation ids are deterministic v5 PKs and the INSERTs have no ON CONFLICT, the re-insert collided → duplicate-key error → aborted txn → (post-#745, where a DB error isn't poison) crash loop on exactly the blocks this PR exists to repair.

**Fix**: dropped AND from_space_id = $4 (and its bind) so the DELETE is scoped by (from_entity_id, type_id, space_id) only. Since we always write from_space_id = space_id, the predicate was redundant for new rows and harmful for old ones. Pre-fix blocks now self-heal on the first recompute (re-acquiring from_space_id = block_space), and the inserts never collide.

**Test gap**: added prefix_null_from_space_projection_self_heals_on_recompute — writes a normal projection, blanks from_space_id to NULL to simulate a pre-fix row (keeping the deterministic id), recomputes the same block, and asserts the recompute succeeds (no duplicate-key) and every row now carries from_space_id = block space. You're right that the existing tests never hit this because they start from a clean DELETE ... WHERE space_id = $1.

ohohoreilly (approved):
LGTM 🚀

PR #758, opened 2026-06-16
feat(api): v2 enriched proposal diff endpoints (folded + media-filtered)

0xNeoDev:
Adds the enriched, context-aware **v2 proposal diff endpoints** and brings them to **`postProcessDiffs` parity** so the geogenesis UI can switch to v2 and delete its client-side `postProcessDiffs` for the proposal path.

- `GET /v2/versioned/proposals/:id/diff`
- `GET /v2/versioned/proposal-groups/diff`

Purely additive — v1 is untouched. Completes Track B of `v2/context.md`.

## What you get
Each **root** entity comes back as a grouped `EntityDiffV2`:
- value changes carry `propertyName`; relation changes carry `typeName` + `toEntityName`, plus inlined `imageUrl`/`videoUrl`
- **block children folded** under `blocks[]` with their own proposed values/relations + data-block config
- **media-property entities** dropped from the top level, URL inlined on the parent relation
- dynamic group keys spread at the entity level

## Pipeline
`resolveRoots` (classify block/media/config children via DB backlinks + edit ops, promote parents incl. orphans to roots) → paginate over **roots** → `buildFoldedPage` (per-block proposed snapshots, config merge, media drop/inline) → shared enrichers (`enrichBlocks`/`enrichNames`/`enrichWithMediaUrls`). Reuses v1 state-construction wholesale (exported additively from `proposal-diff.ts`).

## `postProcessDiffs` parity — all 9 steps covered
| Step | Status |
|---|---|
| classify, orphan-parent backlinks, batch names, synthesize missing blocks, group blocks, apply names | ✅ |
| **data-block config merge** (config on the reified BLOCKS-relation entity, from the edit) | ✅ `resolveRoots` + `buildFoldedPage` merge proposed config; reified entity no longer leaks top-level |
| **media-property filter + inline** (incl. proposed-side URLs) | ✅ |
| **video blocks** (shared `diff.ts` → also fixes the entity endpoint) | ✅ new `videoBlock` type + detection |
| **ranking blocks** → dataBlock | ✅ |
| `imageBlock` via `IMAGE_TYPE` | ✅ |

**Entity-diff endpoint:** inherits video/ranking/image-type from the shared `diff.ts` changes; data-block config was already handled there (committed-vs-committed).

**Accepted divergence:** the orphan new-block *parent-guess heuristic* (frontend L360–394) isn't ported — v2 resolves new blocks from the edit's BLOCKS ops, covering the real case.

## Tests (DB-gated; CI runs `src/versioned/__tests__/` on `api/**`)
Extended `proposal-diff-edit-flow.test.ts`: v1↔v2 parity, name/media enrichment, grouped mode, error parity; **block folded under an orphan parent**, **media dropped + inlined**, **data-block config folded from the reified entity**, **video block folded**, **ranking block folded as dataBlock**, **root-pagination across a page boundary**.

## Verified locally (Postgres container)
**Full versioned suite green (249 tests)**, `tsc --noEmit` clean, `biome check` clean.

## Still out of scope
`GET /v2/versioned/entities/:id` (snapshot) v2 variant — the only remaining endpoint; `postProcessDiffs` doesn't apply to it. Merge into `dev` (feature → `dev`, squash).

### DIFF
--- a/api/drizzle/0064_blocks_to_entity_idx.sql
+++ b/api/drizzle/0064_blocks_to_entity_idx.sql
@@ -1,0 +1,4 @@
+-- Large envs: build CONCURRENTLY first (then the IF NOT EXISTS below no-ops):
+--   SET lock_timeout=0; SET statement_timeout=0;
+--   CREATE INDEX CONCURRENTLY IF NOT EXISTS "relation_versions_blocks_to_entity_idx" ON "relation_versions" USING btree ("to_entity_id","space_id","valid_from_key") WHERE "type_id" = 'beaba5cb-a677-41a8-b353-77030613fc70'::uuid;
+CREATE INDEX IF NOT EXISTS "relation_versions_blocks_to_entity_idx" ON "relation_versions" USING btree ("to_entity_id","space_id","valid_from_key") WHERE "relation_versions"."type_id" = 'beaba5cb-a677-41a8-b353-77030613fc70'::uuid;
--- a/api/drizzle/meta/0064_snapshot.json
+++ b/api/drizzle/meta/0064_snapshot.json
@@ -1,0 +1,3904 @@
+{
+  "id": "e7ef5434-73dd-4eac-af59-3b328c28a2cb",
+  "prevId": "2873933b-353c-4050-9cb9-0c3bb542f723",
+  "version": "7",
+  "dialect": "postgresql",
+  "tables": {
+    "public.app_webhooks": {
+      "name": "app_webhooks",
+      "schema": "",
+      "columns": {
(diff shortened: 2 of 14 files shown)


0xNeoDev:
Follow-up tracked in Linear **PRO-71** — add a `MAX_AFFECTED_ENTITIES` cap to bound per-request work on bulk proposals (the `to_entity_id` partial index in this PR fixes the *latency*; the cap is defense-in-depth against pathological/huge proposals, and applies to v1 too). https://linear.app/geobrowser/issue/PRO-71

PR #769, opened 2026-06-22
chore: dev to main

0xNeoDev:


ohohoreilly (changes requested):
The index design is correct — partial on `type_id = BLOCKS`, columns `(to_entity_id, space_id, valid_from_key)`, well-targeted for `batchGetBlockParentsAtVersion`,   and the `.sql`/snapshot/schema/journal are all consistent. The concern is purely operational: it's a plain `CREATE INDEX` (with `IF NOT EXISTS`), which locks out
  writes on `relation_versions` for the whole build and will stall ingestion on a production-sized table.

  The header comment documents the safe path (manual `CREATE INDEX CONCURRENTLY` first, then the migration no-ops), and since `CONCURRENTLY` can't run inside
  Drizzle's transactional wrapper, that manual pre-build is the right pattern. It just has to actually be on the deploy checklist — otherwise the migration job will
  block writes. Can we confirm that step is part of the rollout?

  ### 🟡 Decide: fix here or fast-follow

  **v2: a media-property child can be silently dropped.** (`api/src/versioned/v2/proposal-diff.ts`, `resolveRoots`)

  Only block-children and config entities promote their parent into `parentRoots`; media children don't. This is fine whenever the edit *creates* the media relation
  (the parent becomes affected), which is what the tests cover. But if an edit only does `updateEntity` on an existing media entity (e.g. changes `IMAGE_URL`) without
  touching the referencing parent, the media entity is classified as a `mediaChild`, dropped, no parent is promoted, and the change disappears from the response.

  Fix mirrors the block path: a media-child→parent backlink that promotes those parents into roots. It's a narrow case, so a tracked fast-follow is acceptable if we
  don't want to hold the promotion — but let's not lose it.

 ### 🟢 Lower severity / confirm

  - **notification-indexer — no blockers.** The deterministic tie-break (`ORDER BY from_entity_id, (space_id = $1) DESC, v.id`) is total and stable; NULL handling and
  the page-entity name resolution are correct; no panics or injection. One thing to confirm: the proposer-name path dropped its `lookup_entity_name` fallback while
  the space-name path kept one — looks intentional (the old fallback only produced a `"Space <uuid>"` placeholder), just want to confirm the asymmetry is deliberate.
  Optional: log the swallowed DB errors (`.ok().flatten()`) so genuine failures aren't indistinguishable from "no name."
  - **v2 minor:** `batchGetMediaUrls`/`...AtVersion` pick an arbitrary row on duplicate `IMAGE_URL` entries (no `ORDER BY`) — worth a deterministic sort, ironically
  the same fix the Rust change in this PR applies. The single-proposal v2 endpoint also emits plural-phrased error messages ("one or more proposals…") where v1 uses
  singular.

  ### Test coverage gaps (v2)

  The folding/active-proposal tests are thorough, but:

  1. **The versioned/historical path is untested** — all fixtures are active proposals, so `baseVersionKey` is always null and none of the new `*AtVersion` query
  functions ever execute. Largest untested surface.
  2. **No grouped-endpoint folding test** (only a smoke test), despite it sharing the fold logic and adding multi-edit op flattening.
  3. **No test for the media-orphan case above** or for deleted/removed blocks. The v1↔v2 parity test passes only because its fixture touches no blocks/media, so it
  doesn't actually validate folding parity.

  ### Action items
  - [x] Confirm the manual `CREATE INDEX CONCURRENTLY` is on the deploy runbook
  - [ ] Decide media-orphan gap: patch here or tracked follow-up + add the missing test
  - [ ] Confirm the notification-indexer proposer-name fallback removal is intentional

ohohoreilly (approved):
LGTM 🚀

0xJagger (approved):

PR #772, opened 2026-06-22
fix(ranking-indexer): make membership recompute idempotent; stop constraint-violation crash loop

ohohoreilly:
## Incident
`KafkaConsumerGroupEmpty` on `ranking-indexer-v1` (prod, ~2026-06-22 20:26 UTC). Consumer group drained to **zero members** — not scaled to 0, but **CrashLoopBackOff**.

## Root cause
A `space.membership` event at **offset 72615** triggered a block re-aggregation. The recompute does DELETE-then-INSERT into `relations`/`values` **without `ON CONFLICT`**. Step-1's value DELETE resolves `entity_id`s through the block's `RANK_POSITION` relations; when a prior value row's relation was missing, the DELETE skipped it and the value INSERT collided on **`values_pkey`**.

The duplicate-key surfaced as `IndexerError::Database`, which `error.rs` classified as **transient** ("writes are idempotent, redelivery converges" — untrue here). It retried 4×, then `std::process::exit(1)` to force redelivery. The offset was never committed, so Kafka redelivered the **same** message → infinite crash loop.

(The `storage.rs` comment already documented this exact failure mode from a prior incident; the earlier fix narrowed the DELETE scope but left the INSERTs non-idempotent.)

## Fix — two layers
1. **Idempotent recompute writes.** `ON CONFLICT (id) DO UPDATE` on the `RANK_POSITION` relations + value rows (the recomputed content wins); `DO NOTHING` on the stable block→submission provenance relations. Re-running converges instead of colliding.
2. **Constraint violations are poison, not transient.** SQLSTATE class `23` (unique / FK / not-null / check) is deterministic and can never converge by retrying — now logged + skipped instead of crash-looping. Genuinely transient DB errors (connection, deadlock `40001`, etc.) still retry as before. This is the safety net so no future non-idempotent write can drain the consumer group again.

## Test
Opt-in e2e regression (`recompute_is_idempotent_when_a_prior_value_row_is_orphaned`) that reproduces the orphaned-value-row state and asserts the recompute returns `Ok` with the projection intact. `cargo build`, `cargo test`, `cargo fmt --check`, `cargo clippy --tests` all clean.

## Deploy note
⚠️ The crashing prod image `1f60e60e` is **not on `main`** — an unmerged/feature build reached production. After merge, redeploy ranking-indexer **from `main`**. If prod is still crash-looping at review time, a stopgap is to roll back to the last-good image `1ce31f55` (06-18) to restore the consumer group while this lands.

### DIFF
--- a/ranking-indexer/src/error.rs
+++ b/ranking-indexer/src/error.rs
@@ -20,12 +20,31 @@ impl IndexerError {
         Self::Decode(msg.into())
     }
 
-    /// True when retrying the same input can never succeed (malformed
-    /// message). Everything else — database, Kafka — is treated as transient:
-    /// retried, and never skipped past, since the writes are idempotent and
-    /// redelivery converges.
+    /// True when retrying the same input can never succeed, so the message
+    /// must be logged + skipped rather than retried (which would crash-loop
+    /// the partition forever, since the offset is never committed).
+    ///
+    /// Two cases are poison:
+    /// - `Decode` — a malformed message; the bytes will never parse.
+    /// - A database **integrity-constraint violation** (SQLSTATE class `23`:
+    ///   unique, foreign-key, not-null, check). These are deterministic — the
+    ///   same input fails identically on every redelivery, so retrying can
+    ///   never converge. The recompute writes are idempotent (`ON CONFLICT`),
+    ///   so we should not hit these in practice; this is the safety net that
+    ///   stops a future non-idempotent write from stalling the consumer group.
+    ///
+    /// Everything else — transient database errors (connection drops,
+    /// deadlocks, serialization failures), Kafka errors — stays transient:
+    /// retried, never skipped, since redelivery can converge.
     pub fn is_poison(&self) -> bool {
-        matches!(self, Self::Decode(_))
+        match self {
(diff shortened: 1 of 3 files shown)


ohohoreilly:
✅ Production recovered. The incident is resolved.

```
 Verification (pod 588c6c457f-4gzp8, image 438337d = the fix):
  - Running 1/1, 0 restarts, up ~90s and counting.
  - Clean startup — created consumer, subscribed to knowledge.edits + space.membership under group ranking-indexer-v1, "ranking-indexer started."
  - No Transient error / exiting / BackOff lines — the exact signature that was repeating every ~7s before.

  The proof it cleared the poison: the old pod crashed within ~7s of every start (4 retries → exit(1)). This one has now survived well past that window with zero
  restarts, which means it processed past offset 72615 and is consuming normally. With an active member back in the group, the KafkaConsumerGroupEmpty alert condition
  is cleared.

  Timeline: alert ~20:26 → diagnosed (poison-message crash loop, non-idempotent recompute) → fix PR #772 → squash-merged (438337d) → auto-deploy → healthy pod at
  20:55.
  ```

PR #789, opened 2026-06-29
feat(api): support returning entities without a value in entities_ordered_by_property

0xJagger:
## Summary

`entitiesOrderedByPropertyConnection` only ever returned entities that **have** a value for the ordered property. The backing function (`entities_ordered_by_property`) drives entirely from the `"values"` table and applies a per-type non-null/non-empty predicate, so entities that match the space/type filters but lack a value for the property were never candidates — e.g. a score-sorted view returned only the handful of entities that happened to have a score.

This adds an opt-in way to include those value-less entities.

## Changes

- **Migration `0067_support_entities_without_value.sql`** — `CREATE OR REPLACE`s the function with a new parameter `include_without_value boolean DEFAULT false`. PostGraphile auto-exposes it as `includeWithoutValue` on both `entitiesOrderedByProperty` and `entitiesOrderedByPropertyConnection`.
- **Integration test** `entitiesOrderedByPropertyWithoutValue.test.ts`.

## Behavior

- **Default `false`** → emits the existing query unchanged (same SQL, same plan). All current callers are byte-for-byte unaffected.
- **`true`** → entities of the given `typeIds` within `spaceIds` that have no usable value are `UNION ALL`'d into the result.
- **A missing value is treated as zero** for numeric types (`integer`/`float`/`decimal`) via `COALESCE(sort_value, 0)`: `DESC` → positives → zero/null → negatives; `ASC` the reverse. Non-numeric types order value-less entities last (`NULLS LAST`).
- The value-less candidate set is anchored on the TYPES relation (entities carry no space/type column), so the flag **requires `typeIds`**; when `true` without `typeIds` it falls back to scored-only output.
- `includeWithoutValue` is nullable at the GraphQL layer; an explicit `null` is coalesced to `false` (does not opt in).

## Testing

- Integration test covers: arg exposure on both fields, default exclusion, ASC/DESC null-as-zero ordering, explicit `null` == `false`, and the no-`typeIds` fallback.
- Verified end-to-end against a fresh `postgres:16` via `scripts/setup-test-db.ts` (all migrations apply clean; tests pass). `tsc` and `biome` clean.

### DIFF
--- a/api/drizzle/0067_support_entities_without_value.sql
+++ b/api/drizzle/0067_support_entities_without_value.sql
@@ -45,6 +45,7 @@ CREATE OR REPLACE FUNCTION public.entities_ordered_by_property(
 )
 RETURNS SETOF entities AS $fn$
 DECLARE
+  types_rel_id  constant uuid := '8f151ba4-de20-4e3c-9cb4-99ddf96f48f1';  -- TYPES relation: links an entity to its type entity
   resolved_type   text;
   sort_expr       text;
   null_pred       text;
@@ -109,7 +110,7 @@ BEGIN
   IF type_ids IS NOT NULL AND cardinality(type_ids) > 0 THEN
     type_pred := format(
       ' AND EXISTS (SELECT 1 FROM relations r WHERE r.from_entity_id = v.entity_id AND r.type_id = %L AND r.to_entity_id = ANY(%L::uuid[]) AND r.space_id = v.space_id)',
-      '8f151ba4-de20-4e3c-9cb4-99ddf96f48f1', type_ids);
+      types_rel_id, type_ids);
   END IF;
 
   -- Value-less entities can only be bounded by the TYPES relation; without type_ids there is
@@ -138,11 +139,9 @@ BEGIN
     END IF;
 
     -- Candidate set for value-less entities: entities of type_ids (within space_ids).
-    -- '8f151ba4-de20-4e3c-9cb4-99ddf96f48f1' is TYPES_RELATION_ID, the system "Types" relation
-    -- that links an entity (from_entity_id) to its type entity (to_entity_id).
     candidate_pred := format(
       'r.type_id = %L AND r.to_entity_id = ANY(%L::uuid[])',
-      '8f151ba4-de20-4e3c-9cb4-99ddf96f48f1', type_ids) || rel_space_pred;
+      types_rel_id, type_ids) || rel_space_pred;
 
     sql :=
          'WITH scored AS ('
--- a/api/drizzle/0067_support_entities_without_value.sql
+++ b/api/drizzle/0067_support_entities_without_value.sql
@@ -59,6 +59,16 @@ DECLARE
   with_unscored   boolean;
   sql             text;
 BEGIN
+  -- include_without_value requires space_ids. The value-less branch enumerates the full
+  -- type population and the whole set is ordered/materialized before PostGraphile applies
+  -- LIMIT/OFFSET (pagination happens outside this function, see 0060), so an unbounded
+  -- (all-spaces) candidate scan could be very large. Requiring space_ids bounds the scan to
+  -- the requested spaces. COALESCE so an explicit NULL flag behaves like false (no guard).
+  IF COALESCE(include_without_value, false) AND (space_ids IS NULL OR cardinality(space_ids) = 0) THEN
+    RAISE EXCEPTION 'space_ids is required when include_without_value is true'
+      USING ERRCODE = '22023';
+  END IF;
+
   IF data_type IS NOT NULL THEN
     resolved_type := lower(data_type);
   ELSE
@@ -128,6 +138,8 @@ BEGIN
     END IF;
 
     -- Candidate set for value-less entities: entities of type_ids (within space_ids).
+    -- '8f151ba4-de20-4e3c-9cb4-99ddf96f48f1' is TYPES_RELATION_ID, the system "Types" relation
+    -- that links an entity (from_entity_id) to its type entity (to_entity_id).
     candidate_pred := format(
       'r.type_id = %L AND r.to_entity_id = ANY(%L::uuid[])',
       '8f151ba4-de20-4e3c-9cb4-99ddf96f48f1', type_ids) || rel_space_pred;
--- a/api/drizzle/0067_support_entities_without_value.sql
+++ b/api/drizzle/0067_support_entities_without_value.sql
@@ -103,8 +103,11 @@ BEGIN
   END IF;
 
   -- Value-less entities can only be bounded by the TYPES relation; without type_ids there is
-  -- no cheap, well-defined candidate set, so fall back to scored-only behavior.
-  with_unscored := include_without_value AND type_ids IS NOT NULL AND cardinality(type_ids) > 0;
+  -- no cheap, well-defined candidate set, so fall back to scored-only behavior. COALESCE guards
+  -- the nullable flag: PostGraphile exposes includeWithoutValue as a nullable Boolean, and a
+  -- NULL must behave like false (NULL AND ... yields NULL, which would otherwise fall through
+  -- the IF NOT below into the unscored branch).
+  with_unscored := COALESCE(include_without_value, false) AND type_ids IS NOT NULL AND cardinality(type_ids) > 0;
 
   IF NOT with_unscored THEN
     -- Backward-compatible path (identical to 0060): only entities WITH a usable value.
--- a/api/drizzle/0067_support_entities_without_value.sql
+++ b/api/drizzle/0067_support_entities_without_value.sql
@@ -1,0 +1,153 @@
+-- Support returning entities WITHOUT a value for the ordered property in
+-- entities_ordered_by_property (and its PostGraphile *Connection field).
+--
+-- Background: the function (see 0060) drives entirely from the "values" table and applies
+-- a per-type NOT NULL / non-empty predicate, so only entities that HAVE a usable value for
+-- `property_id` can ever appear. Entities that match the space/type filters but have no
+-- value for the property were never even candidates (this is the reported bug: a scored
+-- "table view" only returned the handful of entities that happened to have a score).
+--
+-- This adds an opt-in `include_without_value` parameter (default false, so existing queries
+-- are byte-for-byte unchanged — same SQL, same plan). When true, entities of `type_ids`
+-- within `space_ids` that have no usable value for the property are unioned in with a NULL
+-- sort value.
+--
+-- Ordering of the value-less entities (per product guidance: "scores can be negative, so
+-- null should be considered the same as zero"):
+--   * numeric data types (integer/float/decimal): the NULL sort value is ordered as 0 via
+--     COALESCE(sort_value, 0). DESC => positives -> zero/null -> negatives; ASC the reverse.
+--     Entities with a literal 0 and value-less entities sort together, which is intended.
+--   * non-numeric types (text/date/time/datetime/boolean/point): "zero" is undefined, so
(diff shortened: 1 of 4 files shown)


api/drizzle/0067_support_entities_without_value.sql
```diff
@@ -0,0 +1,156 @@
...
+  dir := CASE WHEN sort_direction::text = 'DESC' THEN 'DESC' ELSE 'ASC' END;
+
+  IF space_ids IS NOT NULL AND cardinality(space_ids) > 0 THEN
+    space_pred     := format(' AND v.space_id = ANY(%L::uuid[])', space_ids);
+    rel_space_pred := format(' AND r.space_id = ANY(%L::uuid[])', space_ids);
+  END IF;
+
+  IF type_ids IS NOT NULL AND cardinality(type_ids) > 0 THEN
+    type_pred := format(
+      ' AND EXISTS (SELECT 1 FROM relations r WHERE r.from_entity_id = v.entity_id AND r.type_id = %L AND r.to_entity_id = ANY(%L::uuid[]) AND r.space_id = v.space_id)',
```
ohohoreilly, line 112: **Space-scoping asymmetry: a valued entity can surface as value-less.**

The `scored` CTE's type semi-join here requires the TYPES relation to be in the *same space as the value row* (`r.space_id = v.space_id`), but the value-less `candidates` set below (line 131) requires only that the TYPES relation be in *any* of `space_ids` (`r.space_id = ANY(space_ids)`). These conditions diverge whenever `space_ids` contains 2+ spaces.

Scenario: `space_ids = [A, B]`; entity X has `integer = 100` in space A but its only `TYPES → type` relation is in space B. X fails this semi-join (no type relation in A) and is excluded from `scored`; it *is* a candidate (type relation in B ∈ `space_ids`), and since it's absent from `scored` it lands in `unscored` with `NULL → COALESCE 0`. Net: X sorts at the zero position despite holding value 100 — arguably worse than its prior behavior (absent entirely).

In the single-space case the two conditions coincide, so this is multi-space-only. Worth confirming whether an entity can legitimately be typed and valued in different spaces in practice; if so, consider aligning the candidate predicate with the scored semi-join (or vice-versa).

api/drizzle/0067_support_entities_without_value.sql
```diff
@@ -0,0 +1,156 @@
...
+      '8f151ba4-de20-4e3c-9cb4-99ddf96f48f1', type_ids) || rel_space_pred;
+
+    sql :=
+         'WITH scored AS ('
+      || '  SELECT DISTINCT ON (v.entity_id) v.entity_id, ' || sort_expr || ' AS sort_value'
+      || '  FROM "values" v'
+      || '  WHERE v.property_id = ' || quote_nullable(property_id) || ' AND ' || null_pred || space_pred || type_pred
+      || '  ORDER BY v.entity_id, ' || sort_expr || ' ' || dir
+      || '), candidates AS ('
+      || '  SELECT DISTINCT r.from_entity_id AS entity_id FROM relations r WHERE ' || candidate_pred
```
ohohoreilly, line 153: **Potential large unbounded scan + sort per request on the new path.**

With `include_without_value = true`, `candidates` enumerates *every* entity of `type_ids` within `space_ids`, anti-joins against `scored`, unions, and the entire result is ordered and materialized before PostGraphile applies `LIMIT/OFFSET` outside the function. The `0060` header comment already flags this materialization cost for the smaller scored-only set; the value-less set is strictly larger (potentially the full type population), so for a common/large type a paginated table view now pays a full scan+sort on every page.

If the target types are known-small this is fine — otherwise worth a note like `0060`'s deferred-pagination comment, or an explicit follow-up.

api/src/kg/__tests__/entitiesOrderedByPropertyWithoutValue.test.ts
```diff
@@ -0,0 +1,239 @@
...
+	return (result.data?.entitiesOrderedByProperty ?? []).map((e) => e.id)
+}
+
+const baseVars = {
+	propertyId: undash(PROPERTY_ID),
+	spaceIds: [undash(SPACE)],
+	typeIds: [undash(TYPE_ID)],
+}
+
+describe("entitiesOrderedByProperty — include entities without a value", () => {
```
ohohoreilly, line 171: **Coverage is integer-only.** The test thoroughly exercises `integer` (DESC/ASC zero-positioning, default exclusion, explicit `null`, no-`typeIds` fallback), but the non-numeric branch — `order_expr := 'sub.sort_value ' || dir || ' NULLS LAST'` — is never executed, nor are `decimal`/`float`/`text`. The multi-space layout that triggers the asymmetry noted on the migration is also uncovered (every fixture lives in one `SPACE`). A `text` case and a two-space case would close both the untested branch and that edge case.

ohohoreilly (commented):
Code review of the `entities_ordered_by_property` change. The default-`false` path is byte-for-byte identical to `0060`, so existing callers are unaffected — confirmed by diffing the two functions. Three findings on the new `include_without_value` path, inline. None block the default path; the space-scoping one (on the migration) is the one worth confirming against real data-layout assumptions before merge.

api/drizzle/0067_support_entities_without_value.sql
0xNeoDev: +1, but don't see this happening in practice?

api/drizzle/0067_support_entities_without_value.sql
0xNeoDev: +1

api/drizzle/0067_support_entities_without_value.sql
```diff
@@ -0,0 +1,156 @@
...
+    IF is_numeric THEN
+      order_expr := 'COALESCE(sub.sort_value, 0) ' || dir;
+    ELSE
+      order_expr := 'sub.sort_value ' || dir || ' NULLS LAST';
+    END IF;
+
+    -- Candidate set for value-less entities: entities of type_ids (within space_ids).
+    candidate_pred := format(
+      'r.type_id = %L AND r.to_entity_id = ANY(%L::uuid[])',
+      '8f151ba4-de20-4e3c-9cb4-99ddf96f48f1', type_ids) || rel_space_pred;
```
0xNeoDev: maybe comment for what `8f151ba4-de20-4e3c-9cb4-99ddf96f48f1` is

api/drizzle/0067_support_entities_without_value.sql
0xNeoDev: Could we change `space_ids` from optional to required? Could help limit this to the input spaces if there isn't an easy solution to this.

maybe?
```sql
IF COALESCE(include_without_value, false)
   AND (space_ids IS NULL OR cardinality(space_ids) = 0) THEN
  RAISE EXCEPTION 'space_ids is required when include_without_value is true'
    USING ERRCODE = '22023';
END IF;
```
for backwards compatibility

0xNeoDev (approved):

api/drizzle/0067_support_entities_without_value.sql
ohohoreilly: Likely rare, agreed — it needs someone who knows the data model to confirm whether an entity is ever typed in one space but holds the property value in another (and it only bites on multi-space queries). 

If we want to be defensive without waiting on that: the root cause is that the `unscored` anti-join filters against `scored`, which is type-space-coupled (`r.space_id = v.space_id`). If instead it excluded any entity that has a *usable value for the property within `space_ids`* — independent of where the TYPES relation lives — the leak goes away:

```sql
unscored AS (
  SELECT c.entity_id, NULL::<cast> AS sort_value
  FROM candidates c
  WHERE NOT EXISTS (
    SELECT 1 FROM "values" v
    WHERE v.entity_id = c.entity_id
      AND v.property_id = <property_id>
      AND <null_pred> <space_pred>   -- note: no type_pred here
  )
)
```

That restores consistency with the pre-PR behavior (such an entity is simply absent, as in 0060) rather than surfacing it wrong-sorted at zero. Doesn't fully "fix" the deeper question of showing it *with* its value, but it removes the regression. Happy to push the patch if we decide it's worth it.

api/drizzle/0067_support_entities_without_value.sql
ohohoreilly: > Could we change `space_ids` from optional to required?

Good guard for the scan-size concern — bounds the candidate set, and it's backward-compatible since it only fires on the opt-in path. 👍

One clarification so we don't conflate the two threads: requiring `space_ids` doesn't address the asymmetry in the other thread. That one lives *within* a multi-space `space_ids` — `r.space_id = v.space_id` (scored) vs `r.space_id = ANY(space_ids)` (candidates) still diverge even when `space_ids` is non-empty. So this is the right fix for the unbounded-scan worry, and the anti-join change above is the one for the asymmetry.

api/drizzle/0067_support_entities_without_value.sql
ohohoreilly: +1. That's the TYPES relation type id. It's also referenced bare in the scored `type_pred` (~line 101), so rather than a comment in one spot, a declared constant at the top of the function would cover both call sites:

```sql
DECLARE
  types_rel_id constant uuid := '8f151ba4-de20-4e3c-9cb4-99ddf96f48f1';  -- TYPES relation
```

then pass `types_rel_id` into the `format(... %L ...)` calls. (0060 uses it bare too, so this would be a nice cleanup of both.)

api/drizzle/0067_support_entities_without_value.sql
0xJagger: Confirmed — this is intended, not a bug. For an entity to be scored in a space, both its score (the property value) and its TYPES relation need to be defined on the same `space_id`. So an entity with a value in space A but its TYPES relation only in space B isn't scored in any single space, and treating it as value-less (sorting at zero) is the intended outcome rather than excluding it entirely.

Left the predicates as-is and added a multi-space regression test (`f17aea24`) that pins this behavior: an entity valued in space A but typed in space B sorts at the zero position.

api/drizzle/0067_support_entities_without_value.sql
0xJagger: Good catch — adopted the suggestion in `2d5af1f1`: `space_ids` is now required when `include_without_value` is true (`RAISE … ERRCODE 22023`, COALESCE-guarded so an explicit `null` flag doesn't trip it), which bounds the candidate scan to the requested spaces. Added a comment explaining the materialization cost.

The deeper case — a single large type within a space still pays a full scan+sort per page because `LIMIT/OFFSET` is applied outside the function — is the same limitation `0060` documents and deferred (pushing `LIMIT` inside breaks cursor pagination), so I left it as a documented follow-up rather than solving it here.

api/src/kg/__tests__/entitiesOrderedByPropertyWithoutValue.test.ts
0xJagger: Addressed in `f17aea24`. Added: a `text` case (ASC + DESC) exercising the non-numeric `NULLS LAST` branch, a `float` case for the missing-value-as-zero path, a test for the new `space_ids`-required guard, and a dedicated multi-space `describe` (isolated fixtures) covering the value/type cross-space scenario. Thanks for flagging the single-space blind spot.

api/drizzle/0067_support_entities_without_value.sql
0xJagger: Done in `9807f4d6` — extracted into a named constant rather than a comment: `types_rel_id constant uuid := '8f151ba4-…'`, now used at both call sites (the scored `type_pred` and the `candidate_pred`). Cleaner than the bare literal in two places.

ohohoreilly (approved):
LGTM 🚀

PR #798, opened 2026-07-02
feat(hermes): decode v2 governance Action struct (toAddress + toSpaceId)

0xJagger:
## Summary

Updates the v2 governance `Action` decoding to match the new onchain struct shape. The onchain `Action` gained an explicit space-id target, so its ABI tuple went from the legacy 3 fields to 4:

- **Before:** `(address to, uint256 value, bytes data)`
- **After:** `(address toAddress, bytes16 toSpaceId, uint256 value, bytes data)`

`to_address` is the explicit call target; when it is `address(0)` the target is resolved onchain from `to_space_id` via `SpaceRegistry.spaceIdToAddress` at execution time.

Every place that decodes or constructs this tuple is updated to the 4-field shape, and `calldata` is now read from `fields[3]` instead of `fields[2]`.

## Changes

- **`hermes-schema/proto/governance.proto`** — `ProposalAction`: rename `to` → `to_address`, add `to_space_id = 4`.
- **`hermes-schema/src/pb/governance.rs`** — regenerated prost bindings to match the proto.
- **`hermes-pipeline/src/decode.rs`** — update the `sol!` `Action` struct (`to` → `toAddress` + `bytes16 toSpaceId`), the `ProposalAction` Rust struct, and the ABI decoder to expect a 4-field tuple (value/data read from shifted indices).
- **`hermes-pipeline/src/pipelines/governance.rs`** — propagate `to_address`/`to_space_id` through `parse_proposal_created` and log both.
- **`hermes-substream/src/helpers.rs`** — fix `decode_proposal_actions_inner`, which still assumed the 3-field tuple and pulled calldata from `fields[2]`. With the v2 struct the tuple no longer matched, so `extract_proposal_publish_uris` returned nothing and `map_ipfs_uris` emitted no proposal Publish URIs. Now matches the 4-field tuple and reads calldata from `fields[3]`, mirroring the pipeline decoder.
- **`hermes-relay/src/source/mock_events.rs`** — mock encoder emits the new `toSpaceId` field (zero; mock targets by address).
- **`kg-indexer/src/handlers/governance.rs`** — test helpers updated for the new struct fields.

Clean v2 cutover (fresh deployment), so no legacy 3-field fallback is retained.

## Testing

Existing decode/governance unit tests updated to the 4-field tuple and assertions extended to cover `to_address` / `to_space_id`.

### DIFF
--- a/hermes-pipeline/src/decode.rs
+++ b/hermes-pipeline/src/decode.rs
@@ -361,7 +361,8 @@ sol! {
     /// Action to be executed when a proposal passes.
     #[derive(Debug)]
     struct Action {
-        address to;
+        address toAddress;
+        bytes16 toSpaceId;
         uint256 value;
         bytes data;
     }
@@ -467,10 +468,17 @@ pub struct ProposalCreatedData {
 }
 
 /// A single action in a proposal.
+///
+/// `to_address` + `to_space_id` mirror the onchain `Action` call target: when
+/// `to_address` is non-zero it is the explicit target; when it is `address(0)`
+/// the target is resolved onchain from `to_space_id` via
+/// `SpaceRegistry.spaceIdToAddress` at execution time.
 #[derive(Debug, Clone)]
 pub struct ProposalAction {
-    /// Target contract address (20 bytes).
-    pub to: Vec<u8>,
+    /// Explicit call target (20 bytes); `address(0)` ⇒ resolve via `to_space_id`.
+    pub to_address: Vec<u8>,
+    /// GEO space UUID (16 bytes) used to resolve the target when `to_address` is zero.
+    pub to_space_id: Vec<u8>,
     /// ETH value to send (32 bytes, big-endian).
     pub value: Vec<u8>,
     /// Calldata (function selector + encoded args).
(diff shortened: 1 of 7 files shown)


ohohoreilly:
Closing as superseded — this is now implemented directly on `staging`:

- `d619646` feat(hermes-pipeline): decode v2 governance Action struct (toAddress + toSpaceId) — part of the broader governance v2 migration (#759)
- `52e5cca` fix(hermes-pipeline,hermes-substream): decode legacy 3-tuple proposal actions

Staging's version additionally decodes the legacy 3-tuple as a fallback (needed for reindexing historical proposals), which this PR intentionally didn't retain ("clean v2 cutover, no legacy fallback"). Merging this as-is would conflict with staging and, if force-resolved, would regress that fallback — so closing rather than merging.

Separately (not from this PR, found while reviewing the same area): `proposal-executor/src/contracts.ts`'s `DAOSpaceAbi` still declares the old 3-field `Action` tuple for `getLatestProposalInformation`, while the v2 `IDAOSpace.sol` interface confirms it's now 4-field. Opening a separate PR for that.

PR #842, opened 2026-07-29
fix(hermes-substream): point at the self-owned SpaceRegistry on chain 55516

ohohoreilly:
Step 1 of the registry cutover. **Not deployable on its own** — see Deployment below.

The substream filters Action events **by contract address, not by signature**:

```rust
if log.address() != SPACE_REGISTRY_ADDRESS { ... }
```

So this single constant decides which registry the entire indexing chain observes.

## What changes

`0x364231615dcEA33D13C823b13B449DD9F55381E7` → `0xcf13491802747e759e1bb8e364bc43045398d1dd`

The new registry is the one we own, and the full prod snapshot has been migrated onto it: 1,182 spaces, 123,173 actions, **causal 0 violations, cross-check 1182/1182**. Coverage of the v2 environment is 1,071/1,072.

The old registry is **superseded, not a fallback** — its DAO spaces pin it in initializer storage with no setter (`DAOSpace.sol:95`, no other writer), so they can never be re-pointed, and its owner key is unavailable.

## Verification done

- Byte array checked programmatically against the address: 20 bytes, exact match.
- Recompiled wasm **contains** the new address bytes and **does not contain** the old.

## Deployment — three images, together

The `.spkg` is checked in and copied into images at build time. `hermes-pipeline`, `atlas` **and** `hermes-ipfs-cache` all `COPY` it, and it's loaded at runtime from `hermes-relay/src/config.rs:7`. All three must be rebuilt and redeployed.

This also needs a **reindex**: the new registry's history begins at block **14339**, so `SUBSTREAMS_START_BLOCK` can be set there rather than `0`. Per the fresh-deploy runbook, the kg-indexer Kafka consumer group must be reset to earliest or it resumes mid-stream.

**Sequencing matters.** The executor repoint (`SPACE_REGISTRY_ADDRESS` in `proposal-executor/deployment/v2/cronjob.yaml`) must ship *with* this. Repoint the executor without the indexer and executions succeed on-chain but are invisible to the KG — a silent failure. Space IDs need no change; they were preserved on the new registry.

## Review note

The spkg grew 453,636 → 538,943 bytes. The wasm is 348KB and unchanged in shape; the delta is packer metadata from a newer substreams CLI (v1.20.2) than produced the committed artifact. `specVersion` and package name/version are unchanged (v0.1.0). **Worth confirming the deployed runtime accepts the newer package format before the reindex** — I haven't been able to verify that from here.

Targets `staging`, which is the trunk for the v2 cluster (`main` carries the dead chain-19411 address).


ohohoreilly:
-

PR #873, opened 2026-08-05
ci: run tests and checks on the staging branch

ohohoreilly:
`staging` is the integration branch in practice — every one of the last 15 merged PRs targeted it — but **no workflow referenced it**. So PRs into staging ran zero checks, and roughly 62 commits have merged without an automated test run.

This adds `staging` to the `push` and `pull_request` branch filters of the 31 test/check workflows.

## Scope

The only lines changed are branch filters:

```diff
-    branches: [main, dev]
+    branches: [main, dev, staging]
```

57 such lines across 31 files. Nothing else in any workflow is touched.

**Deliberately excluded:**

| Excluded | Why |
|---|---|
| every `*-deploy*.yml` | so this cannot change where anything deploys |
| `dev-sync-check.yml` | specifically about `dev` |
| `search-admin-build.yml` | publishes images, and has no `pull_request` trigger, so it adds nothing to PR CI |

## Verification

- No deploy workflow modified, and no file outside `.github/workflows/` touched — asserted, not eyeballed.
- All 66 workflows still parse as valid YAML.
- Programmatic re-read of the parsed `on:` blocks confirms `staging` present on `pull_request` for all 31 and on `push` for the 26 that have a push trigger (five are PR-only by design). No deploy workflow gained `staging`.
- 17 of the 31 include `.github/workflows` in their `paths`, so **this PR triggers them itself** — the checks appearing below are the proof the change works. The other 14 are path-filtered to their own service and won't run here.

## One thing to know: cargo-deny

`cargo-deny` is included for parity with main/dev, and it has been failing on its daily schedule for some time. It will now show **red on staging PRs too**.

That is pre-existing and unrelated to whatever a given PR changes. `staging` has no branch protection, so it doesn't block merges. But a check that's permanently red trains people to ignore CI, which defeats the point of this PR — it should be fixed or explicitly ignored as a follow-up rather than left as standing noise.

## Follow-ups this doesn't do

- Reconciling `main` and `staging` (their drizzle chains have forked — `0067`/`0068` mean different migrations on each branch).
- Deciding whether `dev` and the 14 `*-deploy-staging.yml` workflows should still exist; `dev` currently has zero unique commits and is 10 behind main.
- Disabling the 16 production deploy workflows before the legacy cluster is shut down — they all break when it goes away.

### DIFF
--- a/.github/workflows/api-cache-integration-tests.yml
+++ b/.github/workflows/api-cache-integration-tests.yml
@@ -2,7 +2,7 @@ name: API Cache Integration Tests
 
 on:
   push:
-    branches: [main, dev]
+    branches: [main, dev, staging]
     paths:
       - 'api/src/kg/valkeyCache.ts'
       - 'api/src/kg/postgraphile.ts'
@@ -12,7 +12,7 @@ on:
       - 'api/package.json'
       - '.github/workflows/api-cache-integration-tests.yml'
   pull_request:
-    branches: [main, dev]
+    branches: [main, dev, staging]
     paths:
       - 'api/src/kg/valkeyCache.ts'
       - 'api/src/kg/postgraphile.ts'
--- a/.github/workflows/api-check.yml
+++ b/.github/workflows/api-check.yml
@@ -2,11 +2,11 @@ name: API Check
 
 on:
   push:
-    branches: [main, dev]
+    branches: [main, dev, staging]
     paths:
       - 'api/**'
   pull_request:
-    branches: [main, dev]
+    branches: [main, dev, staging]
     paths:
       - 'api/**'
 
--- a/.github/workflows/api-integration-tests.yml
+++ b/.github/workflows/api-integration-tests.yml
@@ -2,13 +2,13 @@ name: API Integration Tests
 
 on:
   push:
-    branches: [main, dev]
+    branches: [main, dev, staging]
     paths:
       - 'api/**'
       - 'api/drizzle/**'
       - '.github/workflows/api-integration-tests.yml'
   pull_request:
-    branches: [main, dev]
+    branches: [main, dev, staging]
     paths:
       - 'api/**'
       - 'api/drizzle/**'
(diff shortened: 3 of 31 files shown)


ohohoreilly:
## It works — and it immediately found a real bug

17 checks ran on this PR. Before it, a PR into `staging` ran **zero**.

Results: mostly green, with two failures. Neither is caused by this diff — 57 lines of `branches:` filters in workflow YAML can't affect a runtime test — but both are worth knowing about.

### 1. `cargo-deny` — pre-existing, flagged in the description

Already red on its daily schedule; now visible on staging PRs too.

### 2. `API Integration Tests` — a genuine bug that has been invisible on staging

`src/versioned/__tests__/proposal-diff-edit-flow.test.ts:577` — the grouped diff endpoint returns **500 instead of 200**.

This workflow has passed on `main` every run (07-16 → 07-27) and fails on `staging` the first time it has ever run there. So one of the ~62 untested commits broke it.

I reproduced it locally and traced the root cause:

```
GET /v2/versioned/proposal-groups/diff
  → QueryError, operation: batchGetProposalsWithPublishActions
  → column p.start_time does not exist
```

It's a **schema skew from the governance-v2 migration**. On `staging`, `0067_governance_v2` moved `start_time` off `proposals`:

- `proposals` columns on staging: `id, space_id, proposed_by, created_at, created_at_block, executed_at, current_version` — no `start_time`
- `start_time` now lives on `proposal_versions` (and `proposals_current`)
- `main`'s `schema.ts` still declares `startTime` on `proposals`, which is why it passes there

The query in `batchGetProposalsWithPublishActions` still selects `p.start_time`. Likely introduced or surfaced by #797 ("report unstarted voting window as active in proposal diff"), which added voting-window logic to this path.

The 500 is also a secondary issue: `QueryError` isn't handled by `mapGroupedProposalError`, so a schema bug surfaces as a generic "Internal server error" instead of anything diagnosable.

**Not fixing it here** — this PR should stay a pure CI change, and choosing between joining `proposal_versions` or reading `proposals_current` is a call for whoever owns the governance-v2 schema move. Worth a separate issue.

This is the same failure class as the known `notification-indexer` skew (querying `proposals.end_time`/`name` after they moved to `proposal_versions`), so it may be worth grepping for other `proposals.`-prefixed column reads that predate the migration.

PR #875, opened 2026-08-05
chore(deps): update anyhow, crossbeam-epoch and ruint for RUSTSEC advisories

ohohoreilly:
`cargo-deny` has been failing on its daily schedule for over a week. It isn't a stale ignore or a yanked crate — there are **three live advisories**, including one that touches on-chain numerics.

| Advisory | Crate | Issue |
|---|---|---|
| [RUSTSEC-2026-0190](https://rustsec.org/advisories/RUSTSEC-2026-0190) | `anyhow 1.0.100` → `1.0.104` | unsoundness in `Error::downcast_mut()` |
| [RUSTSEC-2026-0204](https://rustsec.org/advisories/RUSTSEC-2026-0204) | `crossbeam-epoch 0.9.18` → `0.9.20` | invalid pointer dereference in `fmt::Pointer` |
| [RUSTSEC-2026-0220](https://rustsec.org/advisories/RUSTSEC-2026-0220) | **`ruint 1.17.2` → `1.20.0`** | Uint shift ops: incorrect overflow flags, truncated shift amounts |

The third is the one I'd look at hardest: `ruint` is alloy's big-integer type, so it's on the path for on-chain values. I have not assessed exploitability for our usage — flagging it rather than claiming it's harmless.

All three resolve **within semver**. Lockfile only; no manifest or code changes.

## The `ark-*` additions are not what they look like

`ruint 1.20.0` adds six `ark-*` crates to `Cargo.lock`. I checked rather than assumed: they're optional features that aren't enabled — **zero occurrences in the normal dependency graph, zero compiled**. Not a real supply-chain expansion.

They will show up in cargo-deny's duplicate-version warnings though (`ark-ff` now appears at four versions: 0.3.0, 0.4.2, 0.5.0, 0.6.0 — the first three predate this PR). `multiple-versions = "warn"` in `deny.toml`, so that's noise, not a failure.

## Verification

- `cargo build --workspace --all-targets` → clean
- `cargo test -p atlas --test e2e` → **43 passed**
- Full workspace suite with `--no-fail-fast` (see comment below for totals)

⚠️ **CI coverage on this PR is thin, by design of the path filters.** `Cargo.lock` only matches `cargo-deny`'s paths, so **cargo-deny is the only check that runs** — it's green, which proves the advisories are resolved, but it does *not* prove the workspace still builds and tests with the new versions. That part rests on the local runs above. If you want CI to cover it, the lockfile would need adding to the Rust check workflows' `paths`, which is a separate change.

One honest note on that: my first workspace run showed `atlas::test_e2e_scaling_performance` failing. It asserts a wall-clock ratio (200 nodes vs 50 nodes, must be < 16×) and I had three concurrent cargo builds running. It passes in isolation and in a full unloaded suite run. So: load flake, not this bump — but it's a timing-sensitive test that will flake in CI under contention, which is worth knowing independently of this PR.

`deny.toml` also carries a now-stale ignore (`RUSTSEC-2026-0012`, "no crate matched advisory criteria"). Left alone here to keep this PR to one concern.

## Not fixed here

`deny.toml` states the policy plainly: *"NO RUSTSEC vulnerability is ignored here."* This PR upholds that by upgrading rather than adding ignores.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -713,7 +713,7 @@ version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "40c48f72fd53cd289104fc64099abca73db4166ad86ea0b4341abe65af83dadc"
 dependencies = [
- "windows-sys 0.61.2",
+ "windows-sys 0.60.2",
 ]
 
 [[package]]
@@ -724,14 +724,14 @@ checksum = "291e6a250ff86cd4a820112fb8898808a366d8f9f58ce16d1f538353ad55747d"
 dependencies = [
  "anstyle",
  "once_cell_polyfill",
- "windows-sys 0.61.2",
+ "windows-sys 0.60.2",
 ]
 
 [[package]]
 name = "anyhow"
-version = "1.0.100"
+version = "1.0.104"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a23eb6b1614318a8071c9b2521f36b424b2c83db5eb3a0fead4a6c0809af6e61"
+checksum = "330a5ed07fa54e4702c9d6c4174f74427fc0ef6e214bbd677ae50a5099946470"
 
 [[package]]
 name = "ark-ff"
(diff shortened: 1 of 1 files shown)


ohohoreilly:
## Full workspace verification (the numbers I owed)

`cargo test --workspace --no-fail-fast` on the bumped lockfile:

```
79 test binaries    936 passed    14 failed
```

**All 14 failures are kg-indexer e2e, and neither cause is this PR:**

1. Locally they fail at `kg-indexer/tests/e2e.rs:232` — `env::var("DATABASE_URL").expect("DATABASE_URL must be set")`. No database on my machine; environmental.
2. In CI they fail for a different, pre-existing reason: `column "voting_mode" does not exist` (42703). Governance-v2 moved `voting_mode` to `proposal_versions` and the test still selects it `FROM proposals`. That fails identically on #874 (a whitespace-only PR) and on #872, and was last green on `main` on 2026-07-23.

**No failure anywhere outside kg-indexer e2e.** Atlas passed in this run, confirming the earlier `test_e2e_scaling_performance` failure was load flake from concurrent builds, not the dependency bump.

### Correcting my earlier claim

My first run said "full workspace test suite" — it wasn't. `cargo test --workspace` stops at the first failing target, so it aborted at the atlas flake after 147 tests across 3 binaries. The numbers above are the real ones, with `--no-fail-fast`.

PR #879, opened 2026-08-06
DO NOT MERGE - cache cap was harmful (p99 6.9s->52.4s); see comments

ohohoreilly:
## Problem

The v2 API has been OOMKilling on roughly a 2h cycle since 08-04 — 9 pods, 10–21 restarts each, exit 137. This is what's firing `Api5xxRateHighWARNING`.

Production runs the same image on the **same `limits.memory: 2Gi`** and has been clean for 9 days. The only config difference is the response cache:

| | production | v2 |
|---|---|---|
| `VALKEY_URL` | commented out (`api/k8s/production/api.yaml:134`) | set (`api/k8s/v2/api.yaml:140`) |
| `requests` / `limits` | 768Mi / 2Gi | 768Mi / 2Gi (identical) |

#868 enabled the cache on v2 only, on 08-04.

## Measurement

Per-key `MEMORY USAGE` on the v2 Valkey instance, 2026-08-06:

```
DBSIZE 91, used_memory 15.08 MB
  8.39 MB   1 key
  5.24 MB   1 key
  1.05 MB   3 keys
  <=918 KB  86 keys (most ~17 KB)
```

Two entries hold **13.6 of the 15.08 MB**, and both pass the 15 MB cap comfortably. Re-checked ~90s later, both had `ttl=-2` — they expire and get rewritten constantly.

Valkey is shared across all pods, so each large entry is written once and `JSON.parse`d back into a full object tree on **every hit across 9 readers** — up to ~90 MB transient per pod, per the estimate already in `valkeyCache.ts:40-43`. `ttlPerSchemaCoordinate` (`postgraphile.ts:427`) gives these same queries the *longest* TTL (60s), maximising hits per large entry.

The 15 MB cap was also sized against a 4 Gi pod limit — `"30 MB per concurrent write — 0.7% of the 4 Gi pod limit"`. Both manifests say 2Gi.

## Changes

1. **`MAX_CACHEABLE_BYTES` 15 MB → 1 MB.** Keeps 89 of 91 keys — ~2% of entries, ~90% of bytes. Hit rate comes from repeated small queries, not one-off large ones, so it should be largely preserved. The cache stays on.
2. **`bun --smol`.** JSC grows its heap opportunistically and doesn't reliably size against the cgroup limit, so it can walk past 2Gi before collecting. Independent of the cache theory, cheap to revert.

## What this is not

Neither change is the cause, and I want to be clear about that. Responses are large because `paginationCapPlugin.ts:32-33` caps `first` **per argument** (max 1000, default 100) while nested lists multiply — `entities { valuesList(first: 1000) relationsList(first: 1000) }` is legal at every argument and still hundreds of thousands of rows. A per-response node budget is the real fix and is not in this PR.

Also not here, deliberately: bounding on row count in `shouldCacheResult` before serialization. The check lowered here still runs *after* `JSON.stringify`, so this avoids the ioredis buffer, the network send, and every subsequent GET parse — but not the serialization itself.

## Uncertainty

I can't measure how many of the 1.27M recorded hits land on the two large keys; per-key hit counts aren't exposed without keyspace notifications. If they're disproportionately hot, this costs more hit rate than the key count suggests. It's a one-line revert, so measuring by shipping is cheaper than instrumenting for it.

## Verification

- `bun run ci` (format + lint + tsc) exits 0
- 26 tests pass across `valkeyCache.test.ts` and `responseCacheEmptyResults.test.ts`; they reference `MAX_CACHEABLE_BYTES` symbolically so the boundary cases still hold
- `--smol` flag placement verified against bun 1.3.14

Base is `staging`, which is where v2 deploys from — live image is `0d3e0986`, matching `origin/staging` HEAD.

### DIFF
--- a/api/Dockerfile
+++ b/api/Dockerfile
@@ -28,4 +28,9 @@ USER geo
 
 EXPOSE 3000
 
-CMD ["bun", "run", "main.ts"]
+# --smol makes JSC collect far more eagerly instead of growing the heap
+# opportunistically. Bun does not reliably size its heap against the cgroup
+# limit, so in a 2Gi container (limits.memory in api/k8s/*/api.yaml) it can
+# walk past the limit and be OOMKilled before GC gets serious. Costs some
+# throughput; cheaper than a restart loop.
+CMD ["bun", "--smol", "run", "main.ts"]
--- a/api/Dockerfile
+++ b/api/Dockerfile
@@ -28,9 +28,4 @@ USER geo
 
 EXPOSE 3000
 
-# --smol makes JSC collect far more eagerly instead of growing the heap
-# opportunistically. Bun does not reliably size its heap against the cgroup
-# limit, so in a 2Gi container (limits.memory in api/k8s/*/api.yaml) it can
-# walk past the limit and be OOMKilled before GC gets serious. Costs some
-# throughput; cheaper than a restart loop.
-CMD ["bun", "--smol", "run", "main.ts"]
+CMD ["bun", "run", "main.ts"]
--- a/api/src/kg/valkeyCache.ts
+++ b/api/src/kg/valkeyCache.ts
@@ -16,36 +16,55 @@ export type Cache = {
 }
 
 /**
- * Skip caching responses larger than this threshold. Serializing a large
- * response already allocates a second full copy of the data (JSON.stringify),
- * and handing that string to ioredis allocates a third in its send buffer.
- * For 30+ MB responses, this is the exact allocation spike that OOM'd pods
- * before the pagination cap landed. Skipping the SET entirely avoids the
- * ioredis buffer and the network send — the stringify still happens, but
- * the peak is bounded by one copy instead of three.
+ * Skip caching responses larger than this threshold.
  *
- * 15 MB is chosen to cover the full range of expected cacheable responses:
- *   - the 12 MB Query.spaces response (near-static, high-value to cache)
- *   - 2–15 MB Query.entities* / entitiesOrderedByProperty responses
- *   - common 2–5 MB entity-list shapes
+ * The previous value was 15 MB, sized so that the biggest responses — the
+ * ~12 MB `Query.spaces` list and the 2–15 MB `Query.entities*` lists — stayed
+ * cacheable, on the reasoning that they are expensive and near-static so
+ * caching them is where the win is. Two things were wrong with that.
  *
- * Per-request memory impact stays bounded:
- *   - Single SET peak = JS string (15 MB) + ioredis send buffer (15 MB) ≈
- *     30 MB per concurrent write — 0.7% of the 4 Gi pod limit
- *   - 8-pod cache stampede peaks at ~240 MB aggregate; each pod stays
- *     under 30 MB incremental during the stampede
- *   - SET over in-cluster networking is typically <200 ms; the 3 s
- *     commandTimeout gives ~1.5× headroom for bad network days
+ * First, the arithmetic assumed a 4 Gi pod limit ("30 MB per concurrent
+ * write — 0.7%"). `api/k8s/v2/api.yaml` and `api/k8s/production/api.yaml`
+ * both set `limits.memory: 2Gi`. The headroom was half what the sizing
+ * claimed.
  *
- * GET path on a cache hit parses the full JSON to rebuild the JS object
- * tree — ~100–250 ms CPU and up to ~90 MB transient memory per pod on
- * the biggest entries. For expensive queries this is still a net win over
- * hitting the DB; for cheap queries it can approach DB cost.
+ * Second, and the reason this is being lowered: the write path is the cheap
+ * half. On a hit the GET path parses the full JSON back into a JS object
+ * tree, which for the biggest entries costs ~100–250 ms CPU and up to ~90 MB
+ * transient per pod. Valkey is shared across all API pods, so one writer
+ * produces N readers, and `ttlPerSchemaCoordinate` in postgraphile.ts gives
+ * these exact queries the *longest* TTL (60 s) — the combination maximises
+ * hits per large entry, which is precisely the expensive direction.
  *
- * Responses over 15 MB fall through to the DB on every request rather
- * than cache, bounding the worst-case memory allocation path.
+ * Measured on the v2 cluster 2026-08-06, with the API OOMKilling on a ~2 h
+ * cycle (9 pods, 10–21 restarts each) against a production cluster on the
+ * same 2Gi limit that had run 9 days clean with the cache disabled:
+ *
+ *   DBSIZE 91, used_memory 15.08 MB, distributed as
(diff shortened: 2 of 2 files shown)


ohohoreilly:
## Deployed, partially wrong, corrected

Built `54809575` locally and shipped it to the v2 cluster at 20:35Z (Actions is still down, so this was a manual `docker build` + `kubectl set image`). Findings below; **I've dropped `--smol` from this PR.**

### The restarts are mostly not OOM

Exit 137 with `lastState.reason: Error` is a **liveness kill**, not the OOM killer. `/health/liveness` has `timeoutSeconds: 3`, `failureThreshold: 3` — a pod whose event loop stalls ~30s gets SIGKILLed while memory sits at 400–534 Mi of the 2 Gi limit.

The pre-deploy baseline shows the split, and I misread it originally:

| lastState reason | pods |
|---|---|
| `Error` (liveness kill) | **6 of 9** |
| `OOMKilled` | 3 of 9 |

I described this as "the OOM loop." It's a mixed failure mode and the *dominant* half is liveness timeouts under load. The cache work addresses the minority.

### What's actually stalling the event loop

From a killed pod's `--previous` logs, a burst of expensive queries from one client:

- dozens of `query entity` at `cost: 207` against `threshold: 200`, each `relations(first: 200)` with nested `values(first: 5)` / `relations(first: 5)`
- one `query entities` at **4.38 MB / 2775 ms**
- pool acquires 500–634 ms, though `isPressured: false, isSaturated: false` at 66% utilisation
- all from `3.83.157.90`, `origin: null` — the same unattributed-AWS signature as before

`costLoggerPlugin` **logs** cost 207 > 200 and serves the query anyway. The threshold exists but doesn't enforce.

### Why `--smol` is out

No evidence ever supported it — I added it as a cheap parallel bet. It makes JSC collect more aggressively, spending CPU, and CPU is exactly what's scarce while the event loop is stalling. Wrong lever for the dominant failure mode, and plausibly makes it worse. Removed from the cluster at 20:56Z via a `command` override (faster than a rebuild) and reverted here in `8f32726`.

### What stays

`MAX_CACHEABLE_BYTES` 15 MB → 1 MB. That one has measurements behind it and is working:

| | before | after |
|---|---|---|
| largest key | 8.39 MB | **786 KB** |
| `used_memory` | 15.08 MB | 2.49 MB |
| `keyspace_hits` | 1,265,844 | still climbing |

It genuinely reduces memory pressure and should cut the OOM third. It was never going to fix a CPU-bound liveness timeout, and I shouldn't have implied one change would.

### Follow-ups, none in this PR

1. **Enforce the query cost limit** rather than logging it — the direct fix.
2. **Reconsider the liveness probe.** Killing a saturated pod sheds its capacity onto the survivors, which is how one burst cascades across five pods in 30 seconds. A busy pod is not a dead pod.
3. **Per-response node budget** in `paginationCapPlugin` — `first` is capped per-argument while nested lists multiply.

### Deployment drift — please read before deploying api

The cluster currently runs image `54809575` **plus a `command: ["bun","run","main.ts"]` override** that exists only on the deployment, not in git. Once this merges and a normal build runs, the image CMD is correct and the override should be removed:

```
kubectl -n gaia patch deploy api --type=json -p '[{"op":"remove","path":"/spec/template/spec/containers/0/command"}]'
```

Rollback target throughout: `registry.digitalocean.com/geo/api:0d3e0986`.

ohohoreilly:
## Rolled back. The cache cap made things significantly worse.

Measured on v2 after ~20 minutes on `54809575`:

| | before | on the 1 MB cap |
|---|---|---|
| **p99 latency** | 6.88 s | **52.4 s** |
| **5xx ratio** | 0.04–1.7% | **22.1%** |
| restarts | ~0.8 /pod/hr | ~9 /pod/hr |

Reverted to `0d3e0986` at 21:09Z.

### Why it backfired

The two entries I evicted *were* the expensive queries — `Query.spaces` and `Query.entities`, the ones `ttlPerSchemaCoordinate` deliberately gives a 60 s TTL **because** they're costly. Capping at 1 MB meant they stopped being cached at all, so every request re-ran them against Postgres and re-serialised the result. A 7.6× p99 blowup is exactly what that predicts.

I had the tradeoff backwards. I reasoned "89 of 91 keys retained, so we keep ~98% of the value" — counting *keys*, when the value is concentrated in the two entries that cost the most to recompute. Size and cost correlate here, and the cap selected against precisely the entries worth caching.

### The other error: this was never mainly an OOM problem

Exit 137 with `lastState.reason: Error` is a **liveness kill**, not the OOM killer. Baseline split across 9 pods: **6 `Error`, 3 `OOMKilled`**. I read the 137s as OOMs and built the whole cache theory on the minority failure mode.

### What was actually worth finding

`/health/liveness` had `timeoutSeconds: 3`, `failureThreshold: 3` — ~30 s of stall and the pod is SIGKILLed while using 400–534 Mi of a 2 Gi limit. That kills a pod that is alive and merely busy, which drops its in-flight requests, sheds its load onto the survivors, and brings it back with a cold JIT. One burst from a single client took out five pods in 30 seconds.

Widened to `timeoutSeconds: 10`, `failureThreshold: 6` (~60 s before a kill). Readiness left alone at 5 s / 3, so a saturated pod still leaves the load-balancer rotation gracefully instead of being destroyed. **This is the one change still live on the cluster** and it is not yet in git — it needs a real PR against `api/k8s/v2/api.yaml` and `api/k8s/production/api.yaml`.

### Cluster state

- image `0d3e0986` (original), `command` override removed
- liveness `10s` / `6` retained
- `MAX_CACHEABLE_BYTES` back to 15 MB, cache fully enabled as before

### What this PR should become

Nothing in the current diff should merge. The real work, in order:

1. **Enforce the query cost limit.** `costLoggerPlugin` logs `cost: 207` against `threshold: 200` and serves the query anyway. Bursts of these are what stall the event loop.
2. **Land the liveness probe change in git** so the cluster stops drifting from the manifests.
3. **Per-response node budget** in `paginationCapPlugin` — `first` is capped per *argument* (max 1000, default 100) while nested lists multiply.
4. If the response cache is revisited: bound on **cost**, not bytes. Evicting large-and-cheap is fine; evicting large-and-expensive is what broke this.

ohohoreilly:
Closing — the diff here should not land. Measured on the v2 cluster: the 1 MB cache cap drove p99 from 6.88s to 52.4s and 5xx to 22.1%, because the entries it evicted were precisely the expensive queries the cache existed to serve. Rolled back at 21:09Z; p99 returned to 9.37s against a 9.42s pre-incident baseline.

The findings stay readable in the comments above. The one worthwhile outcome — the liveness probe diagnosis — is now #880.

Real follow-ups, none of which are a revision of this diff:
1. Enforce the query cost limit (costLoggerPlugin logs cost 207 vs threshold 200 and serves anyway)
2. Per-response node budget in paginationCapPlugin
3. If the response cache is revisited: bound on cost, not bytes

PR #888, opened 2026-08-12
fix(atlas): measure the substream stall against the chain tip, not silence

ohohoreilly:
fix(atlas): measure the substream stall against the chain tip, not silence

#887 shipped a stall detector keyed on "no block arrived for 600s". That is
wrong on this chain and would have crash-looped atlas.

Chain 55516 only produces blocks on demand. Measured over a representative
week: ~894 blocks/day, median gap between blocks 2 minutes, p90 8 minutes,
longest genuinely-idle window 130 minutes. That last figure is real chain
idleness, not a broken exporter: chain_tip_block_number and
hermes_latest_processed_block{service="hermes-pipeline"} are independent series
and both sat frozen at 25748 across the window, then advanced to 25752
together.

Against that behaviour the shipped code would have restarted atlas 402 times
that week (57/day), and AtlasStreamStalled would have paged 81 times. The repo
already recorded this hazard, in the directory #887 edited:
chain-tip-exporter's LATEST_BLOCK_BEHIND_THRESHOLD_SECS is 86400 with the
comment "Chain 55516 only produces blocks on demand".

Lag against the tip is cadence-independent and states the actual failure:
blocks exist that we have not processed. An idle chain gives tip == processed
and is silent by construction; the real wedge shows lag growing without bound.

- stall.rs is now a LagTracker: stall iff (tip - processed) > 10 blocks
  sustained for 20 minutes, polling eth_blockNumber every 60s.
- Thresholds come from measured healthy behaviour, not from cadence: a working
  consumer on this chain sits at lag 0 (p99 1, max 8 over the week), and never
  spent more than 2 continuous minutes above lag 1. Substreams delivers at head
  here, so there is no finality offset leaving a healthy consumer behind.
- Everything fails open. An unreadable tip or an unseeded process returns
  Unknown and never restarts; an unreadable tip holds the clock rather than
  clearing it, so a flaky RPC can neither trigger nor mask a stall.
- Progress is seeded from the restored checkpoint. Without that, a restart
  during an idle chain computes the whole chain height as lag and restarts a
  healthy process every 20 minutes.
- Lag is saturating: atlas legitimately observes a block up to 12 ahead of the
  tip poll. That is zero lag, not negative lag.
- RPC_URL is optional, so a deploy landing before the secret key exists
  degrades to the previous behaviour instead of crash-looping.

On the alerting side, #887's other error was excluding atlas from
HermesBehindChainTip, on the grounds that its ~1,750-block replay would trip a
5-block/5-minute threshold. That replay takes 19 seconds, and the rule requires
for: 5m, so the carve-out was unnecessary -- and it removed the one rule that
would have caught this outage correctly, since it is tip-relative and therefore
immune to chain idleness.

- atlas is back in HermesBehindChainTip; that is now the primary signal.
- AtlasStreamStalled (block staleness) and AtlasBehindChainTip (redundant with
  the above) are both deleted.
- AtlasRestartingRepeatedly and AtlasMetricsMissing stay. The latter is the one
  that actually mattered: the two-day silence happened because atlas published
  no metrics at all, so there was nothing for any rule to fire on.

Verification: 13 stall tests, including a regression test that replays the
measured 130-minute idle window and asserts silence. Mutation-tested -- removing
the lag clamp, the unseeded fail-open, the threshold boundary, or the
clock-holding on Unknown each kill exactly one test. 118 atlas lib tests pass,
fmt and clippy clean, all four manifests pass kubectl apply --dry-run=server,
and all three remaining alert expressions parse against live Prometheus with the
join verified to produce one row per service.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1087,6 +1087,7 @@ dependencies = [
  "prost 0.13.5",
  "rand 0.8.5",
  "rdkafka 0.36.2",
+ "reqwest 0.12.28",
  "rustls",
  "serde",
  "serde_json",
--- a/atlas/Cargo.toml
+++ b/atlas/Cargo.toml
@@ -23,6 +23,9 @@ hermes-schema = { path = "../hermes-schema" }
 hex = "0.4"
 prost = "0.13.5"
 rdkafka = { version = "0.36", features = ["cmake-build", "zstd", "ssl"] }
+# Chain-tip polling for the substream stall detector. Mirrors
+# chain-tip-exporter's feature set so both hit the same RPC the same way.
+reqwest = { version = "0.12", default-features = false, features = ["json", "rustls-tls"] }
 rustls = { version = "0.23.36", default-features = false, features = ["ring", "std", "tls12"] }
 tokio = { version = "1.41", features = ["rt-multi-thread", "macros"] }
 anyhow = "1"
--- a/atlas/src/main.rs
+++ b/atlas/src/main.rs
@@ -149,8 +149,15 @@ impl AtlasSink {
                 baseline_force_write_pending,
             }),
             emitter,
+            // Seed from the restored checkpoint. The stall detector measures lag
+            // against the chain tip, so an unseeded restart during an idle chain
+            // would compute the entire chain height as lag and restart a
+            // perfectly healthy process on a loop.
+            progress: match checkpoint_manager.restored_block_number() {
+                Some(block_number) => StreamProgress::seeded(block_number),
+                None => StreamProgress::new(),
+            },
             checkpoint_manager,
-            progress: StreamProgress::new(),
         };
 
         // Force-write on startup: if there's no baseline on disk yet but we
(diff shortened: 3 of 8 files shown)


ohohoreilly:
### cargo-deny is pre-existing, not from this PR

This PR adds a direct `reqwest` dependency, so the failure needed ruling out rather than assuming. It is unrelated:

**1. Identical to the `staging` baseline.** `cargo deny check advisories` run on this branch and on `staging`, same machine, same toolchain:

```
branch (with reqwest)          staging (baseline)
advisories FAILED              advisories FAILED
error[unsound]: LruCache::pop  error[unsound]: LruCache::pop
warning[advisory-not-detected] warning[advisory-not-detected]
warning[yanked]: keccak        warning[yanked]: keccak
warning[yanked]: metrics-util  warning[yanked]: metrics-util
warning[yanked]: metrics       warning[yanked]: metrics
```

**2. `reqwest` does not pull in `lru`.** `cargo tree -p reqwest | grep -c lru` → `0`. The advisory arrives via two pre-existing paths:

```
lru v0.12.5  └── search-indexer
lru v0.16.3  └── alloy-provider ── alloy ── hermes-relay ── atlas / hermes-ipfs-cache / hermes-pipeline
```

**3. The scheduled `Cargo Deny` run at 09:29 UTC today already failed**, hours before this branch existed.

**Supply-chain impact of the new dependency is nil** — the entire `Cargo.lock` diff is one line, because `reqwest` was already in the graph via alloy's transport:

```diff
   "rdkafka 0.36.2",
+  "reqwest 0.12.28",
   "rustls",
```

I kept `reqwest` rather than reaching for `alloy` because it mirrors `chain-tip-exporter`'s JSON-RPC call exactly, and `hermes-relay` exposes no provider or block-number helper to reuse.

The other three checks pass: Build & Lint, Run E2E Tests, Run Unit Tests.

